PR #13, opened 2023-11-06
skip chunking if empty

### LINKED ISSUE #12: chunk_size must not be zero
### problem

i ran a grid search test which used different destination coordinates. the configuration i used didn't include the grid search plugin. the rtree was unable to parse a destination vertex. the only error i received was:

```
[2023-11-06T15:36:58Z INFO  routee_compass::app::compass::compass_app] creating 2 parallel batches across 16 threads to run queries with chunk size 0
thread 'main' panicked at 'chunk_size must not be zero', routee-compass/src/app/compass/compass_app.rs:241:14
```

the error should be that "destination vertex is missing" and it should not blow up the application (error JSON response).

robfitzgerald:
this PR checks if input plugin processing resulted in 0 queries and returns any error responses if this is the case.

Closes #12.


nreinicke (approved):
Nice, looks good

PR #14, opened 2023-11-06
Rjf/crate readmes

### LINKED ISSUE #3: crate readmes need to be github markdown
each crate's Cargo.toml currently sets it's readme field to the "README.md" file at the crate root. those files are also used by rust docs, and extend github markdown with additional features such as links like `crate::app::compass_app::CompassApp`, which cannot be displayed on crates.io, as seen [here](https://crates.io/crates/routee-compass-core).

let's write markdown content for each crate that is restricted to markdown. let's also make it simple, something like [this](https://crates.io/crates/ucd-trie), because the reader of this readme just wants to have high-level information.

robfitzgerald:
this documentation update moves docs.rs-extended markdown files into a "doc.md" file paired with the lib.rs file that references it. a new README.md has been placed at the root of each crate which is used by crates.io.

Closes #3.

### DIFF
--- a/rust/routee-compass-core/README.md
+++ b/rust/routee-compass-core/README.md
@@ -1,43 +1,32 @@
-# Compass Core Crate
+routee-compass-core
+============
+Crate containing core routing modules of RouteE Compass used by all downstream crates. 
 
-Crate containing core routing modules of RouteE Compass used by all downstream crates.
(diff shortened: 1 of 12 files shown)


nreinicke (approved):
Awesome, this looks good

Issue #11, opened 2023-11-03
parameterize energy model at query time


robfitzgerald:
at present, the server is initialized with a single energy model. but energy models have a relatively small footprint in ram. for example, the Camry Coupe model we used in tests is about 420kb on my computer. it seems like it would be simple to load a catalog of models and allow users to pick the model type at query time. this would allow for researchers to look at how sets of queries vary across models, or, a RouteE/HIVE integration to work with heterogeneous fleets.

if a query included a field requesting some model by name:

```JSON
{
  "model_name": "2016_TOYOTA_Camry_4cyl_2WD"
}
```

under-the-hood, there could be something along the lines of a `HashMap<String, SpeedGradeModelRecord>` that would hold the catalog of in-memory models, such as the entire current catalog (.zip) of 60+ trained models we have. we would then have a `SpeedGradeModelRecord` that has the model and metadata:

```rust
struct SpeedGradeModelRecord {
  name: String,
  model: SpeedGradePredictionModel,
  speed_unit: SpeedUnit,
  grade_unit: GradeUnit,
  energy_rate_unit: EnergyRateUnit
}
```

nreinicke:
Yeah this would be a great improvement and shouldn't take a lot of effort to implement.

PR #16, opened 2023-11-06
Allows multiple energy models at query time

nreinicke:
Implements #11 to load multiple energy models at build time and then extract a single model at query time based on a new `"model_name"` key in the query.

This also updates the default OSM energy config to include all the models in the catalog.


rust/routee-compass/src/app/compass/config/traversal_model/speed_grade_energy_model_builder.rs
```diff
@@ -8,9 +9,10 @@ use crate::app::compass::config::{
 };
 use routee_compass_core::model::traversal::traversal_model::TraversalModel;
 use routee_compass_core::util::unit::{
-    DistanceUnit, EnergyRate, EnergyRateUnit, GradeUnit, SpeedUnit, TimeUnit,
+    grade_unit, DistanceUnit, EnergyRate, EnergyRateUnit, GradeUnit, SpeedUnit, TimeUnit,
```
robfitzgerald, line 12: let's remove this weird import i'm guessing rust-analyzer suggested, that looks as though it gets shadowed by variables in the code anyway

robfitzgerald (changes requested):
perfect. such a great new feature and glad it was a reasonable refactor to support it. one itty bitty, kinda silly change requested.

rust/routee-compass/src/app/compass/config/traversal_model/speed_grade_energy_model_builder.rs
nreinicke: Yeah that's a weird one. I just found out that you can do `cargo fix` and it will automagically remove unused import and apply other linting fixes. I just applied it here to pull out some more imports that aren't being used.

robfitzgerald (approved):
cargo fix to the rescue. looks good to me!

PR #18, opened 2023-11-07
update documentation

nreinicke:
Updates the documentation based on some internal reviews:

 - Add badges that indicate which python version is supported.
 - Update the install instructions to prefer conda for hard-to-install dependencies
 - Update the osm example to be more clear and sequential


robfitzgerald (changes requested):
adding a few additional requests/suggestions based on the test group feedback:
- unclear start point in docs/ - maybe we _add_ a /docs/README.md file _with links that work from github.com browser_ that mirrors intro.md (but doesn't replace it, since that is used by jupyter docs and _i believe_ requires different relative paths)
- the added content at the beginning of the notebook is helpful; can we break the content into a few more sections ("run", "analysis") and perhaps a few more lines explaining what's going on there

nreinicke:
okay, just updated to include additional feedback

robfitzgerald (approved):
:clap:

PR #20, opened 2023-11-08
prefer boxed slices over vecs

robfitzgerald:
this PR changes the use of `Vec<T>` for storing graph and traversal model attributes into `Box<[T]>`. this PR has no associated issue, so i'm including some context here.

### problem

reading data from a file in the existing codebase will rely on the underlying `Vec` capacity growth algorithm. when we are loading from very large files, this grows very large, and the final capacity is very likely to be much greater than the `Vec` size. while `Vec` does provide a `reserve` method, it does not remove unused capacity.

if we instead hold this data in a boxed slice of the `Vec` (a `Box<[T]>`), we end up with a data structure where capacity exactly equals length. slices have the same features we are currently requiring from `Vec`.

see stacko discussions on boxed slices [here](https://stackoverflow.com/a/74556233/4803266) and [here](https://stackoverflow.com/a/39693977/4803266).


robfitzgerald:
started a "developer docs" file with the heading "Prefer `Box<[T]>` over `Vec<T>`, wasn't sure where to put the file though so it ended up in the (python) docs/developer directory.

robfitzgerald:
@nreinicke i'm still thinking a little here about how far to go with `Box<[T]>`. it makes a lot of sense for the lookup tables and graph attributes. but my refactor bled into the CompassApp, changing the input + output plugins from `Vec` as well. i think i went too far. maybe we stick to prefering `Box<[T]>` only for collections that can get _very large?_ one drawback with the plugins is that slices aren't Iterators like `Vec`s are, and we want to iterate over the plugins all the time, so we need to keep creating new iterators for the boxed slices.

nreinicke:
Yeah I agree with you that maybe we prefer the box slice for instances where we gain from being exact about memory or instances where we only perform scans or indexing and prefer vectors for the remaining use cases? Otherwise, the changes make sense to me

nreinicke (approved):
Sweet, looks good to me!

PR #21, opened 2023-11-08
trigger build on release tag prefix "py-"

### LINKED ISSUE #6: use github release to trigger pypi, crates releases
taking a look at [polars](https://github.com/pola-rs/polars), they seem to have figured out how to automate releases using a tag scheme where
  - "rs-*": a rust release
  - "py-*" a python release

they pattern-match off of the release name and activate a [release](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release) action based on the language type. this allows the rust and python "version" of polars to move at different speeds.

our current python build action doesn't have this sophistication, but we depend on it for pypi releases, so we cannot use the naming scheme until we sort out how to get the string pattern match working.

robfitzgerald:
this PR sets the python build operation to only happen when the release's tag begins with the string `py-`. this was tested by creating two releases from two "valid" tags. the build process correctly activated from the `py-` tag and didn't run from the `rs-` tag.

Screen Shot 2023-11-08 at 2 00 58 PM

publishing to pypi and crates is done manually; we only use CI to build python wheels for different OSs and python versions, so for this issue, we really only needed to make sure that the python build didn't happen for rust releases.

Closes #6.

### DIFF
--- a/.github/workflows/python-release.yaml
+++ b/.github/workflows/python-release.yaml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   build:
+    if: startsWith(github.ref, 'refs/tags/py-') # 
     name: build py3.${{ matrix.python-version }} on ${{ matrix.platform || matrix.os }}
     strategy:
       fail-fast: false


nreinicke (approved):
Nice! I'm glad this works. Hopefully we can eventually move into using CI to push the package to PyPI but will have to wait at least until maturin releases the next version with the source distribution fix applied.

PR #22, opened 2023-11-08
To Disk Output Plugin

### LINKED ISSUE #19: CompassApp.run option to write responses to file system
when batches are submitted to CompassApp, it keeps all incremental response objects in memory until the batch is completed. for very large batches, this may result in an out-of-memory error. to avoid this, we can immediately write the response to disk and then discard the object from RAM.

implementation will require adding functionality to the rust CompassApp that writes responses to disk and removes the response object. this can be done one of two ways:
- add another run method to CompassApp that appends each response object from a chunk to some output file
- create a "to disk" output plugin _(rjf: i think i prefer this one, it doesn't require changes to the CompassApp Python or Rust API)_
  - perhaps a refactor of OutputPlugins to perform a `flat_map` operation similar to InputPlugins, so we can flatten the result

serializing responses as JSON and appending them to a file is likely simplest if we stick to a newline-delimited JSON format, and if we serialize the JSON as a single row (no newlines) object.

nreinicke:
Adds a new output plugin "to_disk" that writes results to an output file rather than returning them and holding them in memory. 

Updates all output plugins to return a vector of json values rather than a single value and then flattens the result.

When implementing this, I realized our previous convention of using `"_file"` to find input files to normalize was to ambiguous since we don't want to normalize the new key `"output_file"` for this plugin. So, the new proposal is to use the postfix `"_input_file"` to denote a file that should be normalized.

Closes #19

### DIFF
--- a/.github/workflows/python-build-test.yaml
+++ b/.github/workflows/python-build-test.yaml
@@ -7,6 +7,7 @@ on:
       - "docs/**"
       - "**.md"
   pull_request:
+    branches: [ main ]
     paths-ignore:
       - "docs/**"
       - "**.md"
--- a/.github/workflows/python-build-test.yaml
+++ b/.github/workflows/python-build-test.yaml
@@ -7,7 +7,6 @@ on:
       - "docs/**"
       - "**.md"
   pull_request:
-    branches: [ main ]
     paths-ignore:
       - "docs/**"
       - "**.md"
--- a/.github/workflows/rust-build-test.yaml
+++ b/.github/workflows/rust-build-test.yaml
@@ -7,6 +7,7 @@ on:
       - "docs/**"
       - "**.md"
   pull_request:
+    branches: [ main ]
     paths-ignore:
       - "docs/**"
       - "**.md"
--- a/.github/workflows/rust-build-test.yaml
+++ b/.github/workflows/rust-build-test.yaml
@@ -7,7 +7,6 @@ on:
       - "docs/**"
       - "**.md"
   pull_request:
-    branches: [ main ]
     paths-ignore:
       - "docs/**"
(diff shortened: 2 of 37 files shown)


robfitzgerald (changes requested):
overall looks as expected. good catch on removing the file write from CompassApp.run. 

but, this solution could lead to out-of-order writes from multiple threads. i did a quick google search and found [this stacko question](https://stackoverflow.com/questions/65235821/how-do-i-write-to-a-file-from-different-threads-in-rust). i believe the answer has the right idea for us, which is that we need to hold an `Arc<Mutex<File>>` in the output plugin and write to _that_ in the plugin.

edit: i'm actually not sure if i'm correct. does fs::writeln! actually manage concurrency? i'm reading around now...

update: yes, there is [no guarantee](https://doc.rust-lang.org/std/fs/struct.File.html#platform-specific-behavior) of async locks on file writes, so we will need some kind of solution to prevent concurrent out-of-order writes. that might be the example above, or, maybe also a `RwLock<File>` could be a good pick since it handles the lock management under-the-hood.

nreinicke:
> this solution could lead to out-of-order writes from multiple threads. 

yeah good catch, the old solution also had the downside of opening the file for each write whereas your proposed solution just opens it once, much better. Just pushed up a new commit.

nreinicke:
> perhaps we stew on this and think of an easier way to let users specify the output file location that doesn't feel like a hack.

Yeah that makes sense to me, I agree that it presents some weirdness ergonomically and so we should mull over some options to make it easier to use. 

> i also think we should plan to add information to our documentation.

Yeah I'll take a quick pass at just adding that in now so we at least have something.

PR #24, opened 2023-11-09
v0.3.0

nreinicke:
Bumps version to v0.3.0 and adds in support for python 3.12. Also removes optional pip dependencies since we recommend conda for installing them.


robfitzgerald (approved):
:+1:

PR #26, opened 2023-11-09
allow route to be empty

### LINKED ISSUE #25: `apply_output_processing` should accept results that have no routes
the logic for processing a search result currently fails with an error if the route is empty. however, the search logic allows for running queries without destinations that only intend to report search tree results.

robfitzgerald:
this PR makes it so omitting destination_x or destination_y is not an error in CompassApp or in the Rtree InputPlugin.

this may be unexpected behavior; we may want to think about having a flag for CompassApp for users to opt into this so it isn't the default behavior with a flag like `allow_queries_without_destinations: true` or something like that.

Closes #25.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -351,24 +351,21 @@ pub fn apply_output_processing(
         }
         Ok(result) => {
             log::debug!(
-                "completed route for request {}: {} links, {} tree size",
+                "completed search for request {}: {} edges in route, {} in tree",
                 req,
                 result.route.len(),
                 result.tree.len()
             );
 
-            // should be moved into TraversalModel::summary, queries requesting
-            // min spanning tree result will not have a route.
-            let route = result.route.to_vec();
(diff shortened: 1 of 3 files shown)


nreinicke (approved):
This makes sense. It's starting to feel like there should be separate function calls or services for computing a route and computing an isochrone. Both could share the underlying infrastructure but could be explicit about the input requirements.

robfitzgerald:
> It's starting to feel like there should be separate function calls or services for computing a route and computing an isochrone. Both could share the underlying infrastructure but could be explicit about the input requirements.

yea, it could be good idea. perhaps we could solve it by allowing TOML files to "include" other TOML files as dependencies. that way we could provide meaningful defaults for those different use cases without having to impact the code. as i'm currently writing it, the isochrones are a plugin, the isochrone search is a traversal model, so _really_ what makes them different is the configuration.

so if someone can just do

```toml
include = "routes.toml"
```

vs

```toml
include = "isochrones.toml"
```

and we just pre-load that TOML file, it may be easier for users to specify these configurations, again without bleeding into the codebase. 🤔

PR #28, opened 2023-11-10
optional distance tolerance paramter for rtree plugin

robfitzgerald:
this PR adds optional "distance_tolerance" and "distance_unit" parameters for the rtree plugin. there is no associated issue so i'll describe the problem here.

### problem

currently, if the user loads a road network and then submits a query that is well outside of the road network region, the rtree plugin will still find the nearest vertex. this leads to unexpected behavior where matched origins and destinations may be miles from the user-requested locations.

### solution

when the nearest vertex is found, if the optional tolerance is provided, the plugin checks that the vertex found does not exceed the distance tolerance from the query coordinate. if the tolerance is exceeded, the query fails. if no tolerance is provided, then we get the original behavior (tolerance is _infinite_).

i have updated the osm_default_*.toml files to use a tolerance of 200 meters (0.2km). along the way i also fixed the display implementation for Unit types so that they are not enquoted by double-quotes in their string representation.

### example

i have run a national query on a denver road network, where the destination coordinate is in Massachusetts which is 2000+ km outside of the nearest vertex:

```
$ RUST_BACKTRACE=1 RUST_LOG=debug target/release/routee-compass -c ~/dev/nrel/routee/routee-compass-tomtom/compass_tomtom_denver_energy_smartcore.toml /Users/rfitzger/dev/nrel/routee/routee-compass-tomtom/queries/test_national.json 
...
[2023-11-10T19:30:03Z INFO  routee_compass] Query: Object {"destination_name": String("Jack's Abby Craft Lagers"), "destination_x": Number(-71.4133386), "destination_y": Number(42.2803461), "grid_search": Object {"energy_cost_coefficient": Array [Number(0.0), Number(0.25), Number(0.5), Number(0.75), Number(1.0)]}, "model_name": String("2016_TOYOTA_Camry_4cyl_2WD"), "origin_name": String("NREL"), "origin_x": Number(-105.1710052), "origin_y": Number(39.7402804)}
...
[2023-11-10T19:35:06Z ERROR routee_compass] Error: "coord Coord { x: -71.4133386, y: 42.2803461 } nearest vertex coord is Coord { x: -104.541352, y: 40.044087 } which is 2766.833749832952 kilometers away, exceeding the distance tolerance of 0.2 kilometers"

### DIFF
--- a/python/nrel/routee/compass/resources/osm_default_distance.toml
+++ b/python/nrel/routee/compass/resources/osm_default_distance.toml
@@ -12,7 +12,7 @@ distance_unit = "miles"
 [plugin]
 input_plugins = [
     { type = "grid_search" },
-    { type = "vertex_rtree", vertices_input_file = "vertices-compass.csv.gz" },
+    { type = "vertex_rtree", distance_tolerance = 0.2, distance_unit = "kilometers", vertices_input_file = "vertices-compass.csv.gz" },
 ]
 output_plugins = [
     { type = "summary" },
--- a/python/nrel/routee/compass/resources/osm_default_energy.toml
+++ b/python/nrel/routee/compass/resources/osm_default_energy.toml
@@ -8,7 +8,7 @@ verbose = true
 [plugin]
 input_plugins = [
     { type = "grid_search" },
-    { type = "vertex_rtree", vertices_input_file = "vertices-compass.csv.gz" },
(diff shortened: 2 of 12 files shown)


nreinicke (approved):
Nice, this is a great improvement, thanks for adding in

PR #29, opened 2023-11-10
Rjf/easier compass app interactions

robfitzgerald:
this PR makes it easier to work with CompassApp outside of routee-compass for when users want to append new builders to a CompassAppBuilder instance and then use the `fn try_from(pair: (&Config, &CompassAppBuilder))` method on `CompassApp`. to previously do this, i had to duplicate the config reading code and also manually take mutable builder collections, modify them, and build a new CompassAppBuilder to inject builders.


nreinicke (approved):
This makes sense to me and will definitely make it easier to add custom builders. Thanks for adding in!

Issue #2, opened 2023-10-26
Review, modify libs for cargo API compliance


robfitzgerald:
Best practices for library APIs are [well-documented](https://rust-lang.github.io/api-guidelines/about.html). We have done a rough draft of work towards API use-ability for the extension points of compass; a second sweep based on these guidelines would get us up to standards.

PR #30, opened 2023-11-13
Ndr/cleanup

nreinicke:
Address a couple of items:
 - fix homepage link in pyproject.toml
 - fix deploy-docs action to include proper dependencies
 - remove returning the full search tree from OSM defaults
 - run `cargo clippy` and and fix all the errors and warnings that came up

I also updated VSCode to use [clippy](https://doc.rust-lang.org/stable/clippy/usage.html) to check and lint the code instead of the default of `cargo check`. I had to add the following line to my `settings.json` file: 

```json
"rust-analyzer.check.overrideCommand": ["cargo", "clippy", "--workspace", "--message-format=json"] 
```


nreinicke:
Just added in some updates to the docs to point to the [conda-forge package](https://anaconda.org/conda-forge/nrel.routee.compass) for an installation option.

rust/routee-compass-core/src/util/multiset.rs
```diff
@@ -9,13 +9,13 @@ where
     T: Clone + Copy,
 {
     fn from(sets: &'a Vec<Vec<T>>) -> Self {
-        let final_pos: Vec<usize> = sets.into_iter().map(|v| v.len() - 1).collect();
-        let pos: Option<Vec<usize>> = Some(vec![0 as usize; sets.len()]);
-        return MultiSet {
+        let final_pos: Vec<usize> = sets.iter().map(|v| v.len() - 1).collect();
+        let pos: Option<Vec<usize>> = Some(vec![0_usize; sets.len()]);
```
robfitzgerald, line 13: this looks like a typo. is `0_usize` valid rust code?

rust/routee-compass/src/app/compass/compass_json_extensions.rs
```diff
@@ -7,9 +7,9 @@ pub trait CompassJsonExtensions {
 
 impl CompassJsonExtensions for serde_json::Value {
     /// attempts to read queries from the user in the three following ways:
-    ///   1. top-level of JSON is an array -> return it directly
```
robfitzgerald, line 10: victims of find-and-replace on the word "return", can you _return_ these to how they were?

robfitzgerald (changes requested):
some nice and sensible changes in here. i'm sure that we just took care of some of issue #2 in the process.

a silly find-and-replace text fix requested.

also, can you explain to me how we can load nrel.routee.compass now without __init__.py files? shouldn't it just import as `import compass` now? or does the directory position relative to the pyproject.toml file somehow implicitly form package names?

oh, and, why the name change to `graph_struct` for the `graph` module?

nreinicke:
> why the name change to graph_struct for the graph module?

The name change was the the result of this clippy suggestion: https://rust-lang.github.io/rust-clippy/master/index.html#/module_inception

> can you explain to me how we can load nrel.routee.compass now without init.py files? 

This was the result of [PEP 420 ](https://peps.python.org/pep-0420/) that I only recently learned about. Now, namespaces are preferred to be implicit rather than explicit (as they used to be here) and python can understand how to get to the base level package. This became important when I attempted to build a [new namespace package](https://github.nrel.gov/MBAP/routee-internal) that depended on these other namespace packages.

rust/routee-compass-core/src/util/multiset.rs
nreinicke: This was one of the auto clippy fixes ([here's the lint doc](https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_cast)) and I didn't know this but apparently it is [valid rust code](https://doc.rust-lang.org/rust-by-example/types/literals.html#literals). But just putting in `0` would also work since we have an explicit type hint and so maybe I just do that

rust/routee-compass/src/app/compass/compass_json_extensions.rs
nreinicke: whoops! yes good catch

robfitzgerald (approved):
thanks for all of the improvements here! i do think "graph_struct" is weird, maybe what we need is to change the outer module name to "road_network" or something like that, and keep the inner module name "graph"? but i'll leave that to you if you like it, want to leave it or have a different idea.

nreinicke:
> maybe what we need is to change the outer module name to "road_network" or something like that

yeah I like that idea, just pushed up that change

PR #35, opened 2023-11-20
fix error message

robfitzgerald:
quick fix that changes the error message for `get_config_serde`. the current error message seems like a copy and paste typo, it implies the type is supposed to be _string-parseable_, but it really should be _deserializable_.


nreinicke (approved):
yep yep, good catch

PR #36, opened 2023-11-20
list valid module names with unknown module error

robfitzgerald:
more small fries fallout from mep-routee work, this time to list the valid module names when reporting an unknown module name error


nreinicke (approved):
Makes sense and is much more user friendly

PR #43, opened 2023-11-22
only recurse when array element is array or object

### LINKED ISSUE #39: normalize_file_paths will treat array of strings as array of file paths
i have a custom frontier model and it includes an array of strings:

```toml
[frontier]
type = "mep"
time_limit = 40.0
time_index = 2
road_class_input_file = "/Users/rfitzger/dev/nrel/routee/routee-compass-tomtom/data/tomtom_metro_denver_network/edges-road-class-enumerated.txt.gz"
valid_road_classes = ["1", "2", "3", "4", "5", "6"]
```

(this is a wrapper around two other models, one being our built-in road class model)

when i attempt to load this TOML file, i get the following error:

```
Could not find incoming configuration file, tried 1 and configuration/1. Make sure the file exists and that the config key ends with '_input_file'
```

this is because [normalize_file_paths](https://github.com/NREL/routee-compass/blob/main/rust/routee-compass/src/app/compass/config/config_json_extension.rs#L274) assumes all strings are paths and [recurses if it sees an array](https://github.com/NREL/routee-compass/blob/main/rust/routee-compass/src/app/compass/config/config_json_extension.rs#L320).

robfitzgerald:
This PR makes a slight modification to the recursive call when normalizing paths on a configuration file. If an array is encountered, the function only recurses if the element within the array is either a JSON object or JSON array.

Closes #39.

### DIFF
--- a/rust/routee-compass/src/app/compass/config/config_json_extension.rs
+++ b/rust/routee-compass/src/app/compass/config/config_json_extension.rs
@@ -317,7 +317,15 @@ impl ConfigJsonExtensions for serde_json::Value {
             serde_json::Value::Array(arr) => {
                 let mut new_arr = Vec::new();
                 for value in arr.iter() {
-                    new_arr.push(value.normalize_file_paths(root_config_path)?);
+                    match value {
+                        serde_json::Value::Array(_) => {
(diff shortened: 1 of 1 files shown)


nreinicke (approved):
Nice, yes good catch!

PR #44, opened 2023-11-22
Rjf/traversal abstractions clone issue

### LINKED ISSUE #34: Energy traversal service gets cloned
speed_grade_energy_model_builder.rs has a service defined inline with the builder; in total, the chain looks like this:

SpeedGradeEnergyModelBuilder (1)
  |
SpeedGradeModelService (2)
  |
SpeedGradeEnergyModelService (3)
  |
SpeedGradeModel (4)

there's an extra layer there. in the process, we end up cloning SpeedGradeModelService (2) each time we build the SpeedGradeModel, which now means cloning the hashmap of prediction models (cloning the speed lookup table is ok, that is cloning an `Arc`).

the cloning is likely leading to longer runtimes as each query creates a clone. this should be rewritten so that no cloning is needed.

robfitzgerald:
This PR refactors the energy model build process so that all objects in the energy `TraversalModelService` are wrapped in `Arc` references if they are large heap objects, so cloning is not expensive. it cleans up the build removing one layer of indirection along the way. 

`TraversalModelService` (and builder) traits were defined in routee-compass but have been moved to routee-compass-core so that they are available from the routee-compass-powertrain crate. build methods now return a TraversalModelError on failure, since CompassConfigurationError is no longer in scope.

It's probably a good low-hanging fruit task to move the rest of the builder traits to the core crate too.

i attempted to update the doc reference from the routee-compass crate that references TraversalModelBuilder, but i'm worried it won't take, that we can't make cross-crate references like that. we may need to replace those builder links with a direct hyperlink to the routee-compass-core docs at docs.rs.

Closes #34.

### DIFF
--- a/rust/routee-compass-core/src/model/traversal/mod.rs
+++ b/rust/routee-compass-core/src/model/traversal/mod.rs
@@ -2,5 +2,7 @@ pub mod access_result;
 pub mod default;
 pub mod state;
 pub mod traversal_model;
+pub mod traversal_model_builder;
 pub mod traversal_model_error;
+pub mod traversal_model_service;
 pub mod traversal_result;
--- a/rust/routee-compass-core/src/model/traversal/traversal_model_builder.rs
+++ b/rust/routee-compass-core/src/model/traversal/traversal_model_builder.rs
@@ -1,0 +1,25 @@
+use super::{
+    traversal_model_error::TraversalModelError, traversal_model_service::TraversalModelService,
+};
+use std::sync::Arc;
+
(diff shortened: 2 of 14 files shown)


robfitzgerald:
> Perhaps I can merge this into the ndr/phev branch if/when we feel like it's in a good place since there a going to be a couple of merge conflicts related to the changes on the speed grade energy model?

sounds kinda like the same thing as merging this into main now though, right? either way the changes have to get merged into your branch.

nreinicke (approved):
Yeah good point, I'll just merge main into the ndr/phev branch

PR #48, opened 2023-11-28
Rjf/road class rtree

### LINKED ISSUE #40: filter by road classes doesn't sync with the rtree input plugin
_this issue imported from the internal repo_

r-tree pre-processing maps coordinates to vertex ids on the map. if the user chooses a frontier model that restricts some links, the r-tree will not know of these restrictions. as a result, when the road class frontier model ignores TomTom's "restricted links" (road class 7), the denver beer run test case fails, as the origin or destination must have been pegged to vertices that are only connected via restricted links:

```toml
[frontier]
type = "road_class"
road_class_file = "data/tomtom_metro_denver_network/edges-road-class-enumerated.txt.gz"
valid_road_classes = ["1", "2", "3", "4", "5", "6"]  # ignore 7, "restricted road"
```

```
{
        "error": "no path exists between vertices 57209 and 356583",
        "request": {
            "destination_name": "Comrade Brewing Company",
            "destination_vertex": 356583,
            "destination_x": -104.9009913,
            "destination_y": 39.6757025,
            "energy_cost_coefficient": 0.0,
            "origin_name": "NREL",
            "origin_vertex": 57209,
            "origin_x": -105.1710052,
            "origin_y": 39.7402804
        }
    },
```

some possible solutions:

- make an edge-oriented rtree that is built based on the DirectedGraph or otherwise knows the list of valid edges
  - may require passing the DirectedGraph or the SearchApp in the InputPlugin constructor
- promote map matching algorithms from InputPlugin to its own component of SearchApp, where we can build it based on the combination of the frontier model and the directed graph
- ~modify the a star priority values so they are a tuple of (-int(cost * f), -road_class) where f is some factor such as 100~
  - **no longer possible since we removed road class annotations from `Edge`**
  - requires a lower-bound on cost values, which is practical if cost is utility-based
  - cast to integers in order to make an equality check possible
  - follows the road network hierarchy as a tie breaker, using higher-priority links when the costs are equivalent

robfitzgerald:
This PR introduces an edge-oriented rtree input plugin which optionally takes a road class input. it assigns `EdgeId`s instead of `VertexId`s to incoming queries. if the query has a "road_classes" field, it will restrict matching to only edges which have one of the listed classes. when coupled with a road class frontier model with a matching list of valid road classes, we can get the effect of restricting a search by road class.

along the way:
- AsPath<P> arguments in read_utils were passed by reference which is redundant
- added a few geo-related dependencies; didn't need them, could remove them but we probably want them in our stack
- SearchApp runs vertex-oriented search, and if that fails, attempts to run an edge-oriented search
- normalize file paths was still broken, found another way that array contents were being deleted and fixed it

Closes #40.


robfitzgerald:
two examples using the energy-optimal planner, for a comparison of "drive" vs "walk" mode travel, where
- drive: road classes 1-6
- walk: road classes 3-7 (tried 4-7 but that produced no result)

### "drive"

Screen Shot 2023-11-28 at 9 33 13 AM

### "walk"

Screen Shot 2023-11-28 at 9 35 02 AM

rust/routee-compass/src/plugin/input/default/edge_rtree/edge_rtree_input_plugin.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+impl EdgeRtreeInputPlugin {
+    pub fn new(
+        road_class_file: &Path,
+        linestring_file: &Path,
+        tolerance_distance: Option<Distance>,
+        distance_unit: Option<DistanceUnit>,
+    ) -> Result<Self, CompassConfigurationError> {
+        let road_class_lookup: Box<[String]> =
```
nreinicke, line 71: It could be nice to eventually allow arbitrary frontier models here instead of just road class. For example, if we wanted to apply this to heavy duty trucks where there might be restrictions based on vehicle height/weight/length.

nreinicke (approved):
Cool, this makes sense to me and is nice to see the results for using it. I left a couple of comments but nothing to hold this back.

rust/routee-compass/src/app/compass/compass_app.rs
```diff
@@ -258,7 +258,10 @@ impl CompassApp {
         &self,
         query: serde_json::Value,
     ) -> Result<Vec<serde_json::Value>, CompassAppError> {
-        let search_result = self.search_app.run_vertex_oriented(&query);
+        let search_result = self
+            .search_app
+            .run_vertex_oriented(&query)
+            .or_else(|_| self.search_app.run_edge_oriented(&query));
```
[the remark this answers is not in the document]
robfitzgerald: sure, i think it would be a tad less ambiguous in rare circumstances. like we talked about before, perhaps some flags in the TOML that set this or also set "expect route target coordinate" could be part of a user experience improvement for someone down-the-line.

rust/routee-compass/src/plugin/input/default/edge_rtree/edge_rtree_input_plugin.rs
robfitzgerald: yup. i have another thought from doing this. the user should be able to set these requirements at _query time_. the FrontierModel should be 1) parameterized just like TraversalModelService, or 2) should just be rolled in as a sub-component of a TraversalModel and with some kind of build behavior based on query flags/parameters.

PR #50, opened 2023-11-28
Rjf/termination error

### LINKED ISSUE #38: missing termination "type" falls back to default even if other fields are present
running a configuration with the following:

```toml
[termination]
models = [
    { type = "query_runtime", limit = "00:10:00" },
    { type = "solution_size", limit = 30000000 },
]

```

i would expect this to _fail_, since the (required) `type` field is missing, but instead, it runs. that is because the reified final config, after merging with defaults, looks like this:

```toml
[termination]
type = "query_runtime"
limit = "00:02:00"
frequency = 100_000
models = [
    { type = "query_runtime", limit = "00:10:00" },
    { type = "solution_size", limit = 30000000 },
]
```

here, "models" gets ignored, since the "query_runtime" type is set, and it loads correctly since all fields required are present.

to make it fail, i need to specify an invalid type:

```toml
[termination]
type = "pizza"
models = [
    { type = "query_runtime", limit = "00:10:00" },
    { type = "solution_size", limit = 30000000 },
]
```

which fails as expected:

```
[2023-11-21T20:54:02Z ERROR routee_compass] Could not build CompassApp from config file: unknown module pizza for component termination provided by configuration
Error: CompassConfigurationError(UnknownModelNameForComponent("pizza", "termination"))
```

robfitzgerald:
This PR fixes how we report errors from the termination model builder so that, in the case of a combined model, we give the array index of the failing model.

in this way, when one of a few combined models is missing a key, the user understands where to look. for example, this configuration:

```toml
[termination]
type = "combined"
models = [
    { type = "query_runtime", limit = "00:10:00" },
    { type = "solution_size", limit = 30000000 },
]
```

yields this error message:

```
Could not build CompassApp from config file: expected field frequency for component termination.combined[0] provided by configuration
```

and the 0'th model is in fact missing a key.

Closes #38.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -130,7 +130,7 @@ impl TryFrom<(&Config, &CompassAppBuilder)> for CompassApp {
         // build termination model
         let termination_model_json =
             config_json.get_config_section(CompassConfigurationField::Termination)?;
-        let termination_model = TerminationModelBuilder::build(&termination_model_json)?;
+        let termination_model = TerminationModelBuilder::build(&termination_model_json, None)?;
 
         // build graph
         let graph_start = Local::now();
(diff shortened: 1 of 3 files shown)


nreinicke (approved):
Nice, much more user friendly, thanks for updating this.

PR #51, opened 2023-11-28
Rjf/update dependencies

robfitzgerald:
just a quick dependency version bump before we do our next release. ran a test run and saw no change in output.


nreinicke (approved):
Dig it

PR #55, opened 2023-11-30
Ndr/expose fuel conversion

### LINKED ISSUE #46: Update fuel representations
Right now we allow conversion between petro based fuel and kwh, using a fixed value of [33.41](https://github.com/NREL/routee-compass/blob/6f6411bb5da27c13c94bbd92a2a732144ae3dca8/rust/routee-compass-core/src/util/unit/energy_unit.rs#L17). This should be exposed to the [vehicle configuration](https://github.com/NREL/routee-compass/blob/6f6411bb5da27c13c94bbd92a2a732144ae3dca8/python/nrel/routee/compass/resources/osm_default_energy.toml#L26) since different fuels can have a different energy density. While we're at it, we could also add in gallons_diesel as a new energy unit and update our vehicle library for any diesel vehicle to use this new unit (and it's corresponding conversion between diesel and kwh).

nreinicke:
Expose a new "custom_liquid_fuel_to_kwh" parameter to the dual_fuel_vehicle type that will be used to convert the liquid fuel from the charge sustaining model to kwh. If the parameter is not set in the config, the model uses default conversion factors source from: https://epact.energy.gov/fuel-conversion-factors.

Also adds in a new diesel fuel unit.

closes #46

### DIFF
--- a/python/nrel/routee/compass/resources/osm_default_energy.toml
+++ b/python/nrel/routee/compass/resources/osm_default_energy.toml
@@ -272,7 +272,7 @@ model_input_file = "models/2020_Chevrolet_Colorado_2WD_Diesel.bin"
 model_type = "smartcore"
 speed_unit = "miles_per_hour"
 grade_unit = "decimal"
-energy_rate_unit = "gallons_gasoline_per_mile"
+energy_rate_unit = "gallons_diesel_per_mile"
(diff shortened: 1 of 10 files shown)


robfitzgerald (approved):
nice, makes sense. glad we have diesel in the mix too. thanks Nick!

PR #57, opened 2023-11-30
parallelize + show progress for input plugins

### LINKED ISSUE #56: parallelize input plugin operations
i'm running routee-compass with 2200 queries. each has 3 grid search values. as a result, the original JSON values are cloned 6600 times during the grid search plugin. they are also cloned during the rtree plugin operation. apart from some other usual stuff, these queries also have long WKT strings attached to them that are stored for later.

this is taking a real long time, 10+ minutes on my laptop. we can address this by parallelizing the run of input plugins in CompassApp. if kdam::Bar is thread-safe, we can also provide a progress bar for the user.

robfitzgerald:
this PR parallelizes the input plugin operations using the same chunking approach as the search/output plugin parallelization. it also adds a progress bar (using mutex for shared access) to the processing:

```
input plugins:   7%|█████████▆                                                       |  155/2204 [00:17<03:56, 8.67it/s]
````

edit: realized i could also add a progress bar for the searches too.

```
input plugins: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2204/2204 [04:24<00:00, 8.33it/s]
[2023-11-30T21:35:39Z INFO  routee_compass::app::compass::compass_app] creating 12 parallel batches across 16 threads to run queries with chunk size 520
search:   2%|██▇                                                                                                                                             |  118/6240 [00:27<23:40, 4.31it/s]
```

Closes #56.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -20,12 +20,17 @@ use crate::{
 use chrono::{Duration, Local};
 use config::Config;
 use itertools::{Either, Itertools};
+use kdam::{Bar, BarExt};
 use rayon::{current_num_threads, prelude::*};
 use routee_compass_core::{
     algorithm::search::search_algorithm::SearchAlgorithm,
     util::duration_extension::DurationExtension,
 };
-use std::path::{Path, PathBuf};
+use serde_json::Value;
+use std::{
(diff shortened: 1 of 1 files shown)


nreinicke (approved):
Cool, love that we can have progress bars for both of those steps now. Thanks for adding in!

PR #58, opened 2023-12-01
Rjf/fix rtree distance

### LINKED ISSUE #45: PointDistance for RTreeVertex incorrect
for the vertex rtree, we have a distance function

```rust
impl PointDistance for RTreeVertex {
    fn distance_2(&self, point: &Coord) -> f64 {
        let dx = self.x() - point.x;
        let dy = self.y() - point.y;
        dx * dx + dy * dy
    }
}
```

this looks like a quick take on Euclidean distance but missing the square root call at the end (see [here](https://en.wikipedia.org/wiki/Euclidean_distance#Two_dimensions)).

but we can do better, we have an implementation of haversine (routee-compass-core::util::geo::haversine) that can give us this distance value in meters and avoids some skew that comes from using WGS84 values directly in a Euclidean space.

robfitzgerald:
This PR changes the vertex rtree so it uses haversine for nearest neighbor distance. along the way, i separated the 3 struct definitions in the plugin.rs file so they now each have their own file.

Closes #45.

### DIFF
--- a/rust/routee-compass/src/plugin/input/default/vertex_rtree/mod.rs
+++ b/rust/routee-compass/src/plugin/input/default/vertex_rtree/mod.rs
@@ -1,2 +1,4 @@
 pub mod builder;
 pub mod plugin;
+pub mod vertex_r_tree;
+pub mod vertex_r_tree_object;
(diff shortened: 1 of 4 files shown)


robfitzgerald:
it's a mystery to me. i'm out of time for this so i'm going to close this PR, but i also should revise the edge rtree plugin to use squared distance, so i'll push something up for that, and maybe i can get some help on moving these both to squared haversine distance.

PR #60, opened 2023-12-01
Ndr/54 file error msg

### LINKED ISSUE #54: improve error message when malformed/empty input_file is provided
while building a new TOML file, i stubbed out a required input_file:

```TOML
geometry_input_file = ""
```

and forgot to come back and fill it in. when i ran Compass, i got a somewhat cryptic error:

```
Could not find incoming configuration file, tried  and configuration/. Make sure the file exists and that the config key ends with '_input_file'
```

this doesn't tell us what the key was, and so the resulting message is confusing. also, this message is only possible if we already have met it's later requirement of naming, "that the config key ends with '_input_file'", so that is a red herring.

nreinicke:
Fixes #54 by updating the error message for when a file is not found. With an empty file string in the config, the error would look something like:

```console
[2023-12-01T18:03:43Z ERROR routee_compass] Could not build CompassApp from config file:
            Could not find incoming configuration file '' for key 'dummy_input_file'

            Tried:
             - ''
             - '/Users/nreinick/repos/routee-compass-tomtom/'

            Make sure the file exists and that the config key ends with '_input_file'.

Error: CompassConfigurationError(FileNormalizationNotFound("dummy_input_file", "", "/Users/nreinick/repos/routee-compass-tomtom/"))
```

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -101,7 +101,7 @@ impl TryFrom<(&Config, &CompassAppBuilder)> for CompassApp {
         let config_json = config
             .clone()
             .try_deserialize::<serde_json::Value>()?
-            .normalize_file_paths(&root_config_path)?;
+            .normalize_file_paths(&"".to_string(), &root_config_path)?;
 
         let alg_params = config_json.get_config_section(CompassConfigurationField::Algorithm)?;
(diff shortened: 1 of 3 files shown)


robfitzgerald (changes requested):
totes, just what i was thinking. one little weird thing - the line "Make sure the file exists and that the config key ends with '_input_file'." shouldn't exist, because the only way we reach this error message is when the config key in fact does end with '_input_file'. just that and it looks good to me!

PR #61, opened 2023-12-01
switch to Euclidean squared distance

robfitzgerald:
while working on #58, i discovered i had the wrong understanding of the API for rstar::RTree. it turns out the distance metric should be the _squared euclidean distance_. 

our points are in lat/lon, which is a spherical 2D plane aka non-Euclidean. i would assume a haversine distance formula would produce more accurate values. however, after making this change, a simple test case produced an invalid result.

this PR changes the edge rtree distance function to match the vertex rtree distance function, using the squared Euclidean (lat/lon) distance between points. we can have further discussion in #58 as to whether these should be calculated directly as this is or via some transform for accuracy.

### DIFF
--- a/rust/routee-compass/src/plugin/input/default/edge_rtree/edge_rtree_record.rs
+++ b/rust/routee-compass/src/plugin/input/default/edge_rtree/edge_rtree_record.rs
@@ -47,8 +47,15 @@ impl PointDistance for EdgeRtreeRecord {
             .geometry
             .centroid()
             .unwrap_or_else(|| panic!("empty linestring in geometry file"));
-        let distance = haversine::coord_distance_meters(this_point.0, point.0)
-            .unwrap_or(Distance::new(f64::MAX));
-        distance.as_f64()
+        // as noted in the comments for PointDistance, this should return the squared distance.
+        // haversine *should* work but squared haversine in meters is giving weird results for
+        // the vertex rtree plugin, so, i'm reverting this to euclidean for now. -rjf 2023-12-01
+        // let distance = haversine::coord_distance_meters(this_point.0, point.0)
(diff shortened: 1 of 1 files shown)


nreinicke (approved):
Looks good. It is weird that the haversine calculation work in this context..

robfitzgerald:
> It is weird that the haversine calculation work in this context..

no, that's the point, it _doesn't_. but i figured it would, i did something like this:

```rust
let dist = haversine::get_distance_meters(this, that).unwrap_or(Distance::new(f64::MAX)).as_f64();
return dist * dist
```

because i think of 

$dy * dy + dx * dx == dz * dz$ (Pythagoras)

and

$dz ~= haversine(a, b)$

but it didn't. the test case had 3 points, (0,0), (1,1) and (2,2), and matched to query o/d (0.1,0.1) and (1.9,2.1). for some reason, it thought (0,0) was closest to both o and d. even when i swapped the coordinates with _real_ lat/lon coords (the Denver beer run query), it still didn't work, so it wasn't on account of some edge condition in the spherical approximation. so i'm clearly not getting something about this. i mean, ya, the haversine function is on a spherical plane and not strictly Euclidean, but, it's like, approximately so. but enough difference to blow things up? 🤔

nreinicke:
> It is weird that the haversine calculation work in this context..

oops, that should have included the work _doesn't_, my bad!

But, yeah I also tried your branch with actual lat lon values that were very far from each other and the squared haversine distance failed to match the right point which is suspicious..

I suppose another approach would be to project everything into euclidean space for these types of calculations. I wonder if there are any edge cases where we might not return the true nearest neighbor for euclidean distance using lat/lon?

PR #64, opened 2023-12-07
Download ONNX static libraries

nreinicke:
Updates the routee-compass-powertrain crate to download the ONNX static libraries at compile time. This allows us to distribute the source code without also distributing the large libraries. Users who install the pre-built python wheels will not have to download these libraries but if they want to build the code from source, this will allow them to do so with no extra steps. This also sets the `onnx` feature as a default feature in the routee-compass-powertrain crate and uses it in the routee-compass application crate.

I tested this on MacOS Monterey, Windows 11 and Ubuntu 20. 

In general, the ONNX models are resulting in a ~30% speedup for the metro denver test cases.

Also slips in a couple of changes from `cargo clippy --fix`.


robfitzgerald:
cool! first question, what should i do to test this? do i need a fresh install to check that it does what it says it does?

nreinicke:
> what should i do to test this? do i need a fresh install to check that it does what it says it does? 

yeah! I think if you can compile and run the denver onnx config that would be awesome; I'm actually running into a problem right now where I'm getting a certificate error when I try to compile locally and so it would be good to know if you hit a similar issue (it's running and passing tests on GitHub actions and so It feels like a NREL network specific thing)

rust/routee-compass-powertrain/Cargo.toml
```diff
@@ -25,8 +22,16 @@ serde = { workspace = true, features = ["derive"] }
...
 serde_json = { workspace = true }
 ndarray = "0.15"
-ort = { version = "1.15.3", optional = true }
+ort = { version = "2.0.0-alpha.2", optional = true }
 rayon = { workspace = true }
 
 [features]
-onnx = ["ort"]
+default = ["onnx"]
+onnx = ["ort", "ureq", "flate2", "tar", "zip", "sha256"]
```
robfitzgerald, line 30: if the default is onnx, then how do we build this without onnx? is that like `features = []` or something to override the default?

robfitzgerald (commented):
yep, my compile also failed due to the NREL proxy. hopefully proxy configuration is something we can provide to ort when it attempts to download and compile the onnx runtime. here's my error:

```
error: failed to run custom build command for `ort-sys v2.0.0-alpha.2`

Caused by:
  process didn't exit successfully: `/Users/rfitzger/dev/nrel/routee/routee-compass/rust/target/release/build/ort-sys-00e55dcb084c1040/build-script-build` (exit status: 101)
  --- stdout
  [ort] strategy: "unknown"
  cargo:rerun-if-env-changed=ORT_STRATEGY

  --- stderr
  thread 'main' panicked at '[ort] failed to download https://parcel.pyke.io/v2/delivery/ortrs/packages/msort-binary/1.16.3/ortrs-msort_static-v1.16.3-x86_64-apple-darwin.tgz: Transport(Transport { kind: ConnectionFailed, message: Some("tls connection init failed"), url: Some(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("parcel.pyke.io")), port: None, path: "/v2/delivery/ortrs/packages/msort-binary/1.16.3/ortrs-msort_static-v1.16.3-x86_64-apple-darwin.tgz", query: None, fragment: None }), source: Some(Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) }) })', /Users/rfitzger/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ort-sys-2.0.0-alpha.2/build.rs:27:31
```

robfitzgerald:
a quick review of the [ort docs](https://ort.pyke.io/introduction) and issues page doesn't show any discussions about proxies. 

uh, i'm making boatloads and/or pantloads of assumptions here because i haven't really dived into what ort is doing and how you configured it to make onnx runtime builds easier to manage. i'm assuming this is blowing up at the point where our cargo build process is building ort, and ort is building the underlying c library which needs to make an additional download, and at that point, the c build sub-process doesn't have the right configuration to make the HTTP call. is that what's happening? if so, i would have hoped ort would have had a comment on passing a proxy certificate (.pem file), but, that could also possibly be documented by the underlying onnx runtime c library, which may have an environment variable or something like that where we can provide a path to a .pem file or otherwise configure for proxy.

so, i would look into that maybe, but, also, how important is this?

nreinicke:
> how important is this?

good question, maybe not critical right now? The only implication is that you can't build the source code on the NREL VPN but ideally most users are just installing a pre-built python wheel (which gets built outside the network without issue).

nreinicke:
To add a bit more detail to handling the onnx runtime files manually, it would just involve downloaded the static lib for our target and then setting two environment variables.

robfitzgerald:
> for running on HPC, the user would do a pip install nrel.routee.compass which would install a pre-built wheel

OH right, like you said, this isn't a problem on CI. 

> To add a bit more detail to handling the onnx runtime files manually, it would just involve downloading the static lib for our target and then setting two environment variables.

that sounds like a reasonable amount of work for HPC users to deal with, so i agree this works for today.

PR #69, opened 2023-12-08
inject input plugin

robfitzgerald:
this MEP task improvement is so that i can inject something i want to appear in _every_ query for some input file.

i have 12501226 rows of queries covering every US state. i want to keep that dataset small as possible, but i also want to append this hulking JSON value as the `grid_search`: `{"modes":[{"mode":"walk","road_classes":["4","5","6","7"]},{"mode":"bike","road_classes":["4","5","6","7"]},{"mode":"drive","road_classes":["1","2","3","4","5","6"]}]}`.

this plugin allows me to specify some key and value to inject into each input query. usage looks like this:

```toml
[[plugin.input_plugins]]
type = "inject"
key = "grid_search"
value = '{"modes":[{"mode":"walk","road_classes":["4","5","6","7"]},{"mode":"bike","road_classes":["4","5","6","7"]},{"mode":"drive","road_classes":["1","2","3","4","5","6"]}]}'
format = "json"
```

### DIFF
--- a/rust/routee-compass/src/app/compass/config/compass_app_builder.rs
+++ b/rust/routee-compass/src/app/compass/config/compass_app_builder.rs
@@ -15,8 +15,8 @@ use crate::plugin::{
     input::{
         default::{
             edge_rtree::edge_rtree_input_plugin_builder::EdgeRtreeInputPluginBuilder,
-            grid_search::builder::GridSearchBuilder, load_balancer::builder::LoadBalancerBuilder,
-            vertex_rtree::builder::VertexRTreeBuilder,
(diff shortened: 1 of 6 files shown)


nreinicke (approved):
Nice, I wouldn't have anticipated this use case but it definitely makes sense for keeping your query dataset as small as possible.

robfitzgerald:
> I wouldn't have anticipated this use case but it definitely makes sense for keeping your query dataset as small as possible

I had the thought that this is still maybe not the best way to do this. It has to add a big chunk of json to every query. I was thinking of adding an optional JSON argument to the grid search plugin directly to sort of provide this same feature without all of the cloning and whatnot.

PR #72, opened 2023-12-11
Rjf/query time road classes

### LINKED ISSUE #70: query-time road class filter selection
the road class FrontierModel currently is parameterized at configuration time with the set of valid road classes. in order to explore different sets of valid road classes, we can optionally apply road class filtering based on a query `road_classes` argument.

robfitzgerald:
this PR modifies the road class frontier model so that the set of valid road classes can be set at query time.

in order to use the JSON extension methods, the frontier model (and new service) were moved from core into the routee-compass crate.

if users want the original capability (set road classes for all queries), they can add an inject input plugin to their configuration that adds the road_classes key into each query.

Closes #70.


nreinicke (approved):
This all looks good to me

PR #73, opened 2023-12-12
LRU Cache

nreinicke:
After noting that the ONNX models were taking 26 minutes to run for our national beer run, I added in the option to use an LRU Cache for energy models to see if we could cut down on the large run times. I tested this with 10,000 cache entries on the national beer run over 5 parallel queries and noted that the runtime decreased to around 4 minutes per query (the smartcore models saw an increase in runtime from about 3.5 minutes to 4 minutes).

This does convert speed values into kph integers and grade values into milli integers. This might be too strong an assumption but was made considering that our current data sources typically store speed and grade at this resolution anyways.

The default behavior is to not use any caching but if the key `cache_size` is present in the vehicle configuration, the model will build an LRU cache.

### DIFF
--- a/rust/routee-compass-powertrain/Cargo.toml
+++ b/rust/routee-compass-powertrain/Cargo.toml
@@ -23,15 +23,9 @@ serde_repr = "0.1"
 serde_json = { workspace = true }
 ndarray = "0.15"
 ort = { version = "2.0.0-alpha.2", optional = true }
+lru = { version = "0.12" }
 rayon = { workspace = true }
 
 [features]
 default = ["onnx"]
-onnx = ["ort", "ureq", "flate2", "tar", "zip", "sha256"]
-
-[build-dependencies]
-sha256 = { version = "1.4", optional = true }
-ureq = { version = "2.9", optional = true }
-flate2 = { workspace = true, optional = true }
-tar = { version = "0.4", optional = true }
-zip = { version = "0.6", optional = true }
+onnx = ["ort"]
--- a/rust/routee-compass-powertrain/Cargo.toml
+++ b/rust/routee-compass-powertrain/Cargo.toml
@@ -23,7 +23,6 @@ serde_repr = "0.1"
 serde_json = { workspace = true }
 ndarray = "0.15"
 ort = { version = "2.0.0-alpha.2", optional = true }
-lru = { version = "0.12" }
 rayon = { workspace = true }
 
 [features]
--- a/rust/routee-compass-powertrain/src/routee/prediction/prediction_model_record.rs
+++ b/rust/routee-compass-powertrain/src/routee/prediction/prediction_model_record.rs
@@ -1,14 +1,13 @@
-use std::{
-    num::NonZeroUsize,
-    sync::{Arc, Mutex},
-};
+use std::sync::Arc;
 
-use lru::LruCache;
 use routee_compass_core::{
     model::traversal::traversal_model_error::TraversalModelError,
-    util::unit::{
-        as_f64::AsF64, Distance, DistanceUnit, Energy, EnergyRate, EnergyRateUnit, EnergyUnit,
-        Grade, GradeUnit, Speed, SpeedUnit,
+    util::{
+        cache_policy::float_cache_policy::FloatCachePolicy,
+        unit::{
+            as_f64::AsF64, Distance, DistanceUnit, Energy, EnergyRate, EnergyRateUnit, EnergyUnit,
+            Grade, GradeUnit, Speed, SpeedUnit,
+        },
     },
 };
 
@@ -23,45 +22,10 @@ pub struct PredictionModelRecord {
     pub energy_rate_unit: EnergyRateUnit,
     pub ideal_energy_rate: EnergyRate,
     pub real_world_energy_adjustment: f64,
-    cache: Option<Mutex<LruCache<(i32, i32), EnergyRate>>>,
+    pub cache: Option<FloatCachePolicy>,
 }
 
 impl PredictionModelRecord {
-    #[allow(clippy::too_many_arguments)]
-    pub fn new(
-        name: String,
-        prediction_model: Arc<dyn PredictionModel>,
-        model_type: ModelType,
-        speed_unit: SpeedUnit,
-        grade_unit: GradeUnit,
-        energy_rate_unit: EnergyRateUnit,
-        ideal_energy_rate: EnergyRate,
-        real_world_energy_adjustment: f64,
-        max_cache_size: Option<usize>,
-    ) -> Result<Self, TraversalModelError> {
-        let cache = match max_cache_size {
-            Some(s) => {
-                let size = NonZeroUsize::new(s).ok_or(TraversalModelError::BuildError(
-                    "maximum_cache_size must be greater than 0".to_string(),
-                ))?;
-                let cache = LruCache::new(size);
-                Some(Mutex::new(cache))
-            }
-            None => None,
-        };
-
-        Ok(Self {
-            name,
-            prediction_model,
-            model_type,
-            speed_unit,
-            grade_unit,
-            energy_rate_unit,
-            ideal_energy_rate,
-            real_world_energy_adjustment,
-            cache,
-        })
-    }
     pub fn predict(
         &self,
         speed: (Speed, SpeedUnit),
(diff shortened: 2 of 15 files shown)


rust/routee-compass-powertrain/src/routee/prediction/prediction_model_record.rs
```diff
@@ -19,17 +23,83 @@ pub struct PredictionModelRecord {
...
+                // convert speed to kph and then round to nearest integer
+                let speed_kph_int = speed
+                    .1
+                    .convert(speed.0, SpeedUnit::KilometersPerHour)
+                    .as_f64()
+                    .round() as i32;
+                let grade_millis_int =
+                    grade.1.convert(grade.0, GradeUnit::Millis).as_f64().round() as i32;
+
+                let mut cache = cache.lock().unwrap();
```
robfitzgerald, line 84: handle this result instead of unwrap

rust/routee-compass-powertrain/src/routee/prediction/prediction_model_record.rs
```diff
@@ -19,17 +23,83 @@ pub struct PredictionModelRecord {
...
+        speed_unit: SpeedUnit,
+        grade_unit: GradeUnit,
+        energy_rate_unit: EnergyRateUnit,
+        ideal_energy_rate: EnergyRate,
+        real_world_energy_adjustment: f64,
+        max_cache_size: Option<usize>,
+    ) -> Result<Self, TraversalModelError> {
+        let cache = match max_cache_size {
+            Some(s) => {
+                let size = NonZeroUsize::new(s).ok_or(TraversalModelError::BuildError(
```
robfitzgerald, line 44: oh sweet! i didn't know about NonZeroUsize, that's great!

robfitzgerald (changes requested):
smart thinking. i can imagine this will help a lot as parallelism increases for national runs. there's an unwrap that snuck its way into the PR and i also have a suggestion. instead of caching floats at an implicit numeric precision, why not make that a parameter here. perhaps instead of `cache_size: Option<usize>`, the optional configuration argument is instead an `cache_policy: Option<DecimalCachePolicy>` which could be something like this:

```rust
#[derive(Deserialize, Serialize)]
pub struct DecimalCachePolicy<T> {
  cache: Mutex<LruCache<T>>,
  factor: f64
}

impl DecimalCachePolicy {
  pub fn new(cache_size: usize, precision: i64) -> Result<DecimalCachePolicy, Error> {}

  pub fn check_cache<T>(&self, key: T) -> Result<Option<f64>, Error> {}

  pub fn update_cache<T>(&self, key: T, value: f64) -> Result<(), Error> {}
}
```

this would 
- make it so we can use this elsewhere in the code
- make it so the rounding factor is explicit (so we don't forget it)
- encapsulate the mutex lock/unlock operations instead of exposing them to the prediction model

rust/routee-compass-powertrain/src/routee/prediction/prediction_model_record.rs
nreinicke: oops, yep good catch

nreinicke:
> perhaps instead of cache_size: Option<usize>, the optional configuration argument is instead an cache_policy: Option<DecimalCachePolicy> 

Yeah this is a great idea, I'll take a pass at that. Eventually it might be nice to extend the types of caches we might use too. For example, in large parallel runs it could be advantageous to use a more concurrent cache like [moka](https://github.com/moka-rs/moka) since this approach requires an exclusive lock for all reads and writes (I did test the moka library for the national beer run with 5 parallel processes and it actually bumped up the individual query runtimes from 4 minutes to 8 minutes so maybe it's only more performant with more parallel read/write attempts?)

robfitzgerald:
> For example, in large parallel runs it could be advantageous to use a more concurrent cache like [moka](https://github.com/moka-rs/moka) since this approach requires an exclusive lock for all reads and writes

right. what you added is great for now, but you're right, if we have the time to benchmark cache performance across a variety of inputs (avg trip distance, number of processes, network size, prediction model features, precision), it could be useful for fine-tuning to compare different cache libraries, since there is no one-size-fits-all cache solution. but i think we can probably get a lot of mileage out of this first effort.

nreinicke:
@robfitzgerald - I've updated this based on your suggestion and so now using this might look something like this:

```toml
[[traversal.vehicles]]
name = "2016_TOYOTA_Camry_4cyl_2WD"
type = "single_fuel"
model_input_file = "data/2016_TOYOTA_Camry_4cyl_2WD.onnx"
model_type = "onnx"
speed_unit = "miles_per_hour"
grade_unit = "decimal"
energy_rate_unit = "gallons_gasoline_per_mile"
ideal_energy_rate = 0.02857142857
real_world_energy_adjustment = 1.166
[traversal.vehicles.float_cache_policy]
cache_size = 10_000
key_precisions = [
  0, # speed goes from 71.2 to 71
  3, # grade goes from 0.123 (decimal) to 123 or 123 (millis) to 123000
]
```

robfitzgerald:
heya, just gave this a spin, have a few comments here. 

### caching changes the result

as we were worried, using the cache gives different results. used the denver beer run, onnx configuration, and your precision arguments above. in this case, it doesn't seem to change the final route, but the energy estimate was changed for this 100% energy-optimal route. i can't say that it wouldn't change the route for other inputs.

attr | no cache | cache
--- | --- | ---
edge count | 506 | 506
runtime | 1.331s | 0.363s
distance (miles) | 17.520403978295114 | 17.520403978295114
time (minutes) | 24.676013964317605 | 24.676013964317605
energy (gallons gasoline) | 0.5636607318624347 | 0.562071475961919

unsure what to suggest here, maybe it's worth exploring the precision arguments a little to see if you can get a smaller error? or just push this with the caveat that the user should explore those tradeoffs?

### precision can be negative

and per our conversation, the precision argument should be i64 to allow for negative values. leave it to you whether that is something to roll into this PR or add as an issue for future work.

nreinicke:
Yeah thanks for testing this. I was able to get the denver beer run to report the same cache and no cache energy results with some tweaking on the precisions:

```toml
[traversal.vehicles.float_cache_policy]
cache_size = 10_000
key_precisions = [
    2, # speed goes from 71.23 to 7123
    4, # grade goes from 0.123 (decimal) to 1230 or 123 (millis) to 1230000
]
```

But, it would probably still be good to mark this with the caveat that it _could_ have side effects with respect to the actual results.

> the precision argument should be i64 to allow for negative values.

I played around with this a bit and I think the current implementation does allow for negative values: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb969d13022df63be77c193beb95d903

nreinicke:
OH, you must be talking about letting the precision go in either direction like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a746d087cbdae1eba90b62eac2e518a0

That makes more sense and I think I'll just plop that in

robfitzgerald:
> OH, you must be talking about letting the precision go in either direction 

yes. like take for instance if we have energy in joules. 1 kwh becomes 3,600,000. we may want a precision of -4 to cache this value as 360.

> But, it would probably still be good to mark this with the caveat that it could have side effects with respect to the actual results

ya. where does that caveat go? this is a parameter for prediction models so i guess it would go ... [here?](https://github.com/NREL/routee-compass/blob/main/docs/config.md#energy-model)

nreinicke:
Okay, I've allowed precision to be negative and I've added a note into the config section for the vehicle models with some disclaimers about using the caching feature.

rust/routee-compass-powertrain/Cargo.toml
```diff
@@ -23,15 +23,9 @@ serde_repr = "0.1"
 serde_json = { workspace = true }
 ndarray = "0.15"
 ort = { version = "2.0.0-alpha.2", optional = true }
+lru = { version = "0.12" }
```
robfitzgerald, line 26: not sure we need this import here

robfitzgerald (approved):
looks great. thanks for including some tests of the cache policy and documentation on the feature.

rust/routee-compass-powertrain/Cargo.toml
nreinicke: yep, good catch

PR #74, opened 2023-12-12
document builder>service>model convention

robfitzgerald:
just a simple addition to our documentation.


nreinicke (approved):
Awesome, this is great to include, thanks for adding it in

PR #76, opened 2023-12-14
Add battery electric vehicle model

nreinicke:
Adds in a new battery electric vehicle implementation that keeps track of it's battery energy. This implementation also captures regen braking by offsetting the energy cost by the vehicle battery size so the costs will never be negative but any links for which the vehicle reports negative energy (regen) will be preferred over links in which the vehicle reports positive energy. 

I tested this out using the 2017 Chevrolet Bolt model on the internal denver metro map and got the following result (green is least energy and blue is least time) for a route between Boulder and Building 16:

image

Least Time Summary:

```json
{
  "distance": 30.687880672396663,
  "distance_unit": "miles",
  "time": 31.75302379499293,
  "time_unit": "minutes",
  "vehicle": {
    "battery_energy": 12.845941287871266,
    "battery_soc_percent": 78.5900978535478
  },
  "vehicle_info": {
    "battery_energy_unit": "kilowatt_hours"
  }
}
```

Least Energy Summary

```json
{
  "distance": 29.249235551148406,
  "distance_unit": "miles",
  "time": 84.56192603544797,
  "time_unit": "minutes",
  "vehicle": {
    "battery_energy": 7.885298236178546,
    "battery_soc_percent": 86.85783627303569
  },
  "vehicle_info": {
    "battery_energy_unit": "kilowatt_hours"
  }
}
```

The energy optimal route is taking a serious time penalty to conserve a significant amount of energy which could be somewhat realistic.

This also highlight some other items that need attention.

The least energy route is taking a restricted route through the NREL Flatiron Campus:

image

These types of restrictions should be captured in our frontier model.

The least energy route also takes a hiking trail along table mountain:

image

These should at least be restricted in the frontier model for driving trips and perhaps it would be nice to have functionality to optionally restrict loading of certain links if we know we're going to use the graph exclusively for a particular mode (drive in this case).


rust/routee-compass-powertrain/src/routee/vehicle/default/bev.rs
```diff
@@ -0,0 +1,304 @@
...
+        &self,
+        query: &serde_json::Value,
+    ) -> Result<Arc<dyn VehicleType>, TraversalModelError> {
+        let starting_soc_percent = match query.get("starting_soc_percent".to_string()) {
+            Some(soc_string) => soc_string.as_f64().ok_or(TraversalModelError::BuildError(
+                "Expected 'starting_soc_percent' value to be numeric".to_string(),
+            ))?,
+            None => 100.0,
+        };
+        if !(0.0..=100.0).contains(&starting_soc_percent) {
```
robfitzgerald, line 117: OHHH you _can_ do inclusive ranges in rust. cool!

robfitzgerald (approved):
looks nice. i like that we have firmed up our types for vehicles into ICE, PHEV, and BEV class implementations. thanks for providing some tests as well.

Issue #79, opened 2023-12-14
over-use of cloning on things that could be dereferenced or simply passed by reference


robfitzgerald:
noting that 
- there are quite a few places where something like a newtype is being cloned just because the function signature expects a type passed by value, not by reference (see the units libraries like Distance and DistanceUnit, etc)
- some things are being passed by value that could be passed by reference, and those calls end up cloning those values (can't think of where i've seen this but i know there's a good deal of it)

this unnecessary cloning adds a time and space complexity hit when we are talking about low-level operations such as unit conversion, serialization, or anything in the loop of search, so we should go spelunking to remove them.

PR #81, opened 2023-12-14
normalize energy based on vehicle specific parameters

nreinicke:
Adds in an optional parameter to a vehicle to indicate the maximum expected energy delta on a link. Then, with this value, each vehicle will normalize their energy values to be between 0-1. Each vehicle does this differently. For example, the BEV will offset the energy value first and then normalize the energy based on that new offset value to make sure the result is never negative. The time values are also normalized to be between 0-1 by using 30 minutes as the maximum link time delta.

When running a sweep over 6 energy cost coefficients from 0 to 1, I'm starting to see some intermediate routes (yellow being shortest time and blue being least energy):

image

image

### DIFF
--- a/rust/routee-compass-core/src/util/unit/energy.rs
+++ b/rust/routee-compass-core/src/util/unit/energy.rs
@@ -1,6 +1,7 @@
-use derive_more::{Add, Div, Mul, Neg, Sub, Sum};
+use derive_more::{Add, Mul, Neg, Sub, Sum};
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
+use std::ops::Div;
 use std::{cmp::Ordering, fmt::Display};
 
 use super::{
@@ -21,7 +22,6 @@ use super::{
     Add,
     Sub,
     Mul,
-    Div,
     Sum,
     Neg,
 )]
@@ -38,6 +38,14 @@ impl From<(EnergyRate, Distance)> for Energy {
         let (energy_rate, distance) = value;
         let energy_value = energy_rate.as_f64() * distance.as_f64();
         Energy::new(energy_value)
+    }
+}
+
+impl Div for Energy {
+    type Output = Energy;
+    fn div(self, rhs: Self) -> Self::Output {
+        let value = self.0 / rhs.0;
+        Energy(value)
     }
 }
 
--- a/rust/routee-compass-core/src/util/unit/time_unit.rs
+++ b/rust/routee-compass-core/src/util/unit/time_unit.rs
@@ -3,7 +3,7 @@ use crate::util::serde_ops::string_deserialize;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
 #[serde(rename_all = "snake_case")]
 pub enum TimeUnit {
     Hours,
--- a/rust/routee-compass-core/src/util/unit/time.rs
+++ b/rust/routee-compass-core/src/util/unit/time.rs
@@ -1,6 +1,7 @@
-use derive_more::{Add, Div, Mul, Neg, Sub, Sum};
+use derive_more::{Add, Mul, Neg, Sub, Sum};
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
+use std::ops::Div;
 use std::{cmp::Ordering, fmt::Display};
 
 use super::{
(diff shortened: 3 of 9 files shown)


nreinicke:
The last plot has an interesting pareto front:

image

The point that lies way off the front is the yellow route where the time is close to the optimal time but the energy is way higher that the shortest time route.

nreinicke:
Hmmmm, there's something not quite right with this. Here's a sample run where the energy cost coefficient of 1.0 actually produces a route with more energy than the route with the coefficient of 0.0:

query:
```json
{'model_name': '2017_CHEVROLET_Bolt',
 'starting_soc_percent': 100,
 'grid_search': {'energy_cost_coefficient': [0.0,
   0.2,
   0.4,
   0.6000000000000001,
   0.8,
   1.0]},
 'destination_y': 39.96400556338435,
 'destination_x': -105.09385002667304,
 'origin_y': 39.62257436349091,
 'origin_x': -104.88747463756835}
```

plot:
image

table:
image

This might need a bit more scrutiny before we merge.

rust/routee-compass-core/src/util/unit/energy.rs
```diff
@@ -41,6 +41,14 @@ impl From<(EnergyRate, Distance)> for Energy {
     }
 }
 
+impl Div for Energy {
```
robfitzgerald, line 44: what was wrong with derive_more::Div?

rust/routee-compass-core/src/util/unit/time_unit.rs
```diff
@@ -3,7 +3,7 @@ use crate::util::serde_ops::string_deserialize;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
```
robfitzgerald, line 6: the real problem here is that these convert ops should take references to their arguments, addressed in #79

robfitzgerald (commented):
this is an interesting experiment and i'm glad it confirms to us that there is a way to get a pareto front of alternatives if we tweak the cost function. but i agree, i don't think these changes should be integrated. even if we provide a cost function like this, i think it shouldn't be hard-coded but instead provided as an abstraction, so we can swap it out with alternatives via configuration.

rust/routee-compass-core/src/util/unit/energy.rs
nreinicke: Trying to divide an energy value by another energy value led to this compiler error:

```console
cannot divide `ordered_float::OrderedFloat<f64>` by `routee_compass_core::util::unit::Energy`
the trait `std::ops::Div<routee_compass_core::util::unit::Energy>` is not implemented for `ordered_float::OrderedFloat<f64>`
the following other types implement trait `std::ops::Div<Rhs>`:
  <ordered_float::OrderedFloat<T> as std::ops::Div>
  <ordered_float::OrderedFloat<T> as std::ops::Div<T>>
  <ordered_float::OrderedFloat<T> as std::ops::Div<&'a T>>
  <ordered_float::OrderedFloat<T> as std::ops::Div<&'a ordered_float::OrderedFloat<T>>>
  <&'a ordered_float::OrderedFloat<T> as std::ops::Div<ordered_float::OrderedFloat<T>>>
  <&'a ordered_float::OrderedFloat<T> as std::ops::Div<T>>
  <&'a ordered_float::OrderedFloat<T> as std::ops::Div<&'a T>>
required for `routee_compass_core::util::unit::Energy` to implement `std::ops::Div`
```

rust/routee-compass-core/src/util/unit/time_unit.rs
nreinicke: Yeah that's a good point

nreinicke:
Okay, this is interesting.. if I remove the energy offset from the BEV and just pin negative values to zero energy, I get this result:

image

image

image

So, offsetting the energy values isn't working as I expected, we might want to revisit how we capture regen braking

robfitzgerald:
> we might want to revisit how we capture regen braking

right, we may have rolled in those changes and the VehicleType abstraction without proper vetting. maybe we need to write some tests for the different vehicle types and their helper functions to confirm things are doing what we think they are, and then, also, that they interact with the TraversalModel correctly as well.

robfitzgerald:
> if I remove the energy offset from the BEV and just pin negative values to zero energy, I get this result

actually, this again points to how we need to enforce monotonicity in the cost space. there actually may be no error in the BEV code here, but just finally how you are turning that into costs. i'm free for a quick chat if you want to talk about this.

robfitzgerald:
started #83 to talk about cost function abstractions

nreinicke:
Closing this in favor of moving in the direction proposed in #83

PR #82, opened 2023-12-14
create tree multipoint

robfitzgerald:
a small change here. this function fell out of MEP work and may as well go alongside the create_tree_multilinestring operation.


nreinicke (approved):
Cool, looks good to me

PR #85, opened 2023-12-18
relax JSON string arg types

### LINKED ISSUE #79: over-use of cloning on things that could be dereferenced or simply passed by reference

robfitzgerald:
### edit

this PR modified the signature of methods in `config_json_extension.rs` so that string arguments are _references_ and they can be `String` or `str` types. this makes it so cloning of key names is no longer required for using these methods and at the same time, allows `'static str` arguments, which greatly reduces boilerplate for some applications.

### original post

Nick - started this work to relax what we can pass as arguments to the JSON getter extensions we have. this
1. removes the need to `.clone()` keys a lot
2. relax the types we can pass as arguments, so that &str and &String arguments are both valid:

```rust
let parent_1 = String::from("iteration termination function");
let x = config.get_config_i64(&"limit", &parent_1);
```

just letting you know about it, it's going to touch a lot of files, just making sure this doesn't mess with anything you are up to this morning.

Closes #79.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -105,7 +105,7 @@ impl TryFrom<(&Config, &CompassAppBuilder)> for CompassApp {
         let config_json = config
             .clone()
             .try_deserialize::<serde_json::Value>()?
-            .normalize_file_paths("", &root_config_path)?;
+            .normalize_file_paths(&"", &root_config_path)?;
 
         let alg_params = config_json.get_config_section(CompassConfigurationField::Algorithm)?;
         let search_algorithm = SearchAlgorithm::try_from(&alg_params)?;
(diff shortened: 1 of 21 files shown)


robfitzgerald:
@nreinicke mkay this one is ready. 

i was able to get the arguments to be flexible by making them `&dyn AsRef<str>`, which both String and str can be. the implementation of AsRef<str> for String becomes
```rust
impl AsRef<str> for String {
    #[inline]
    fn as_ref(&self) -> &str {
        self
    }
}
``` 

so i think this _should_ be a zero-cost abstraction for both types.

nreinicke (approved):
Nice, this really cleans things up! I especially like that we can do &"config_key" rather than String::from("config_key"), much simpler!

robfitzgerald:
totes! also, another thing i slipped in was an implementation of `AsRef<str>` for `CompassConfigurationField`, so that those can also be passed directly. if there are other enums out there that become JSON key lookup values, we can create AsRef<str> instances for those to get that same behavior.

PR #86, opened 2023-12-19
bump version to 0.5.0

nreinicke:
Bumps version to 0.5.0 in anticipation of a new release


robfitzgerald (approved):
we need a cool naming ritual for our releases like we had/have with hive

nreinicke:
> we need a cool naming ritual for our releases like we had/have with hive

yeah we definitely do, maybe next release we can launch it

PR #87, opened 2023-12-21
Rjf/cost abstractions

### LINKED ISSUE #83: CostFunction: abstraction or convention
aggregating ~arbitrary~ raw state transition data into a cost function is a naive way to estimate the cost of graph traversal and doesn't correctly capture the tradeoffs. we see this in our results, where a simple cost function such as

$$\alpha * t + (1-\alpha) * e$$

where $t$ is the time to traverse a link, $e$ is the energy to traverse a link, and $\alpha$ is a energy/time cost coefficient in the range $[0,1]$.

but since the magnitudes of $T$ and $E$ are different, and $E$ space may not increase monotonically, the composition of these values lacks the invariants of a ~good~ valid cost function for a graph search.

### what we want

we want to combine state transitions in a way that is properly normalized. in this task, we are talking about the software engineering task of designing _how_ we model the function from state transition to `Cost` in code. this could be an abstraction or maybe just a convention, because
1. while an abstraction feels right, it may not be helpful since any `CostFunction` is tightly coupled with it's `TraversalModel` that defines what the state transition variables are
2. but the right abstraction would let us define in one place some aggregation logic that could be applied to arbitrary `TraversalModel`s

### some ideas

here's an idea that doesn't try to solve 100% of the problem but come up with something reasonable to configure. we could provide cost mappings by unit type in configuration of the traversal model. for each unit type in compass, we could have an optional cost mapping:

```rust
pub enum CostMapping {
  Raw,
  Scalar { value: f64 },
  // leaving room for extension if we need to do any fancier maths, maybe not needed
  Poly2 { x0: f64, x1: f64 }, 
  Exp { base: f64, exp_coefficient: f64 },
  Combined(Vec<CostMapping>),
  ...
}

pub struct CostMappingConfiguration {
  distance: Option<CostMapping>,
  time: Option<CostMapping>,
  energy: Option<CostMapping>,
  // ... future units here
}

impl CostMappingConfiguration {
  pub fn distance_cost(&self) -> Result<Option<Cost>, Error> {} 
  // ... etc
}
```

and users could specify that in configuration, assumed to match units in the traversal model:

```toml
[traversal.cost]
distance_cost = { type = "scalar", value = 0.0123 }  # some mileage-based cost factor
time_cost = { type = "scalar", value = 0.0456 }  # some value-of-time cost factor
energy_cost = { type = "scalar", value = 0.00789 } # some price per KwH, possibly based on charging
aggregation = "sum"
```

the user could provide a cost for any unit type, and then the traversal model scoops up from this whatever cost mapping it needs.

the only potential problem here is that the mapping is 1:1 with the unit type. that _may_ be too simple for users who may want to come up with a mix of costs. or it may be fine, they may be able to pre-compute how those factors reduce to a single factor, or they may be able to make use of the Combined approach for that too.

robfitzgerald:
this PR splits out cost estimation from the `TraversalModel` into a new long running service that builds `CostModel` instances. cost defaults can be set at app configuration time, but also can be overridden at query time. specific state variables are assigned cost rates by name. each state variable name can also be assigned a cost coefficient value. this involved a pretty large refactor of the library. the notebook example has been updated to use the cost model, using some reasonable defaults documented in the default TOML files.

### how to use costs

- for each type of search state variable there may be an instance of a cost rate which maps the state attribute to a cost value
- in order to consider a state variable as part of the cost function, it must appear amongst the `state_variable_coefficients` query field
- the state variable coefficients are used to scale costs that are used _during the search_
- the final search state is used to compute the total cost, using any state variable appearing in the `state_variable_coefficients` field
- query overwrites function differently depending on the field

field | type | description | overwrite behavior
--- | --- | --- | ---
state_variable_coefficients | `HashMap<String, f64>` | lists the state variables to use and a search objective coefficient for each | replace
vehicle_state_variable_rates | `HashMap<String, VehicleCostRate>` | list vehicle state attributes and the rate we use to compute their cost | merge
network_state_variable_rates | `HashMap<String, NetworkCostRate>` | list network state attributes and the lookup method used to compute their cost | merge _(not yet implemented)_
cost_aggregation | `CostAggregation` | how `&[(&String, Cost)]` becomes the final `Cost` | replace

### example configuration

for example, the configuration:

```toml
[cost]
cost_aggregation = "sum"
[cost.state_variable_coefficients]
distance = 0
time = 1
[cost.vehicle_state_variable_rates.time]
type = "raw"
```

is used for the end-to-end CompassApp speeds test, which uses raw time costs with a coefficient of 1 (100% of time considered). 

a default cost model has been added to the config.default.toml file to cover distance and time values.

### new APIs

the different rate types are found in vehicle_cost_rate.rs. the default vehicle rate configuration is to use the raw values for any vehicle costs:

```toml
[cost.vehicle_state_variable_rates.distance]
type = "raw"
[cost.vehicle_state_variable_rates.time]
type = "raw"
[cost.vehicle_state_variable_rates.energy]
type = "raw"
[cost.vehicle_state_variable_rates.energy_liquid]
type = "raw"
[cost.vehicle_state_variable_rates.energy_electric]
type = "raw"
```

there is also a key for defining network-based cost rates:

```toml
[cost.network_state_variable_rates.parking]
type = "traversal_cost"
cost_input_file = "parking-data.csv.gz"
```

those variants are found in network_cost_rate.rs. this feature is un-tested.

Closes #83.


nreinicke:
This looks awesome! I've been doing some simple testing this morning and noticed that the runtimes are quite a bit longer for the raw energy optimal routes. For example, I used the [compass_tomtom_denver_energy_smartcore.toml ](https://github.nrel.gov/MBAP/routee-compass-tomtom/blob/main/compass_tomtom_denver_energy_smartcore.toml)config and the following query:

```json
{
  "model_name": "2017_CHEVROLET_Bolt",
  "starting_soc_percent": 100,
  "destination_y": 39.62627481432341,
  "destination_x": -104.99460207519721,
  "origin_y": 39.798311884359094,
  "origin_x": -104.86796368632217,
  "state_variable_coefficients": {
    "time": 0,
    "energy": 1
  }
}
```

For the code from the main branch, this runs in 500ms or so but on this branch we time out after 2 minutes.

Does the query look right to you? I've confirmed that the model is using the default raw configuration based on the results from the shortest time route ("time": 1, "energy": 0) which gives me the cost summary:

```json
{
  "cost": {
    "energy": 8.111696303262868,
    "time": 25.139387990744755,
    "total_cost": 33.251084294007626
  },
  "info": {
    "cost_aggregation": "sum",
    "network_state_variable_rates": {},
    "state_variable_coefficients": {
      "energy": 0,
      "time": 1
    },
    "state_variable_indices": [
      [
        "time",
        1
      ],
      [
        "energy",
        2
      ]
    ],
    "vehicle_state_variable_rates": {
      "distance": {
        "type": "raw"
      },
      "energy": {
        "type": "raw"
      },
      "energy_electric": {
        "type": "raw"
      },
      "energy_liquid": {
        "type": "raw"
      },
      "time": {
        "type": "raw"
      }
    }
  }
}
```

Maybe the cost estimate is off for raw energy?

robfitzgerald:
> Does the query look right to you?

interesting. wondering 
  1. how many edges were explored (tree_size)
  2. what the traversal summary looks like

if the cost function isn't correct for the cost estimate, we may have lost our a* heuristic behavior that we want.

nreinicke:
Yeah let me bump up the time limit and run again to get the summary. I did notice that the final energy value of `8.11` from the shortest time search does match what we get from the version on the main branch, making me thing the actual traversal cost for energy is still working the same.

nreinicke:
Okay, this is a new one. I raised the time limit but ended up with this error instead after 3 minutes:

```json
{
  "error": "loop in search result revisits edge 23718",
  "request": {
    "destination_edge": 84955,
    "destination_x": -104.9946020751972,
    "destination_y": 39.62627481432341,
    "model_name": "2017_CHEVROLET_Bolt",
    "origin_edge": 510159,
    "origin_x": -104.86796368632216,
    "origin_y": 39.798311884359094,
    "query_weight_estimate": 21.983725812182445,
    "road_classes": [
      "1",
      "2",
      "3",
      "4",
      "5",
      "6"
    ],
    "starting_soc_percent": 100,
    "state_variable_coefficients": {
      "energy": 1,
      "time": 0
    }
  }
}
```

Do we still have a catch for negative energies?

robfitzgerald:
> Do we still have a catch for negative energies?

hmm good point. was that happening in the VehicleType? if that's the case, i didn't touch it, but if it was in the energy traversal model, i may have wiped it out by mistake. also, i should throw in a minimum cost in the CostModel so we can't have costs of zero or less. i'll go digging, thanks for your help investigating!

rust/routee-compass-core/src/model/cost/cost_configuration.rs
```diff
@@ -0,0 +1,9 @@
+use super::vehicle::vehicle_cost_mapping::VehicleCostMapping;
+
+pub struct CostConfiguration {
```
nreinicke, line 3: Looks like this might be an artifact of a previous iteration as I can't find it linked in anywhere

rust/routee-compass-core/src/model/cost/vehicle/vehicle_cost_rate.rs
```diff
@@ -0,0 +1,50 @@
+use crate::model::{traversal::state::state_variable::StateVar, unit::as_f64::AsF64, unit::Cost};
+use serde::{Deserialize, Serialize};
+/// a mapping for how to transform vehicle state values into a Cost.
+/// mappings can be a single instance of Raw, Factor, or Offset mapping.
+///
+/// when multiple mappings are specified they are applied sequentially (in user-defined order)
+/// to the state value.
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum VehicleCostRate {
```
nreinicke, line 10: I think the word "rate" is a little confusing to me here since Raw and Offset are not actually a rate. I wonder if we could find a more general work that captures the idea that it transforms the cost in some way (maybe `VehicleCostTransform` or `VehicleCostOperation`?)

rust/routee-compass-powertrain/src/routee/vehicle/default/bev.rs
```diff
@@ -46,12 +45,16 @@ impl VehicleType for BEV {
     fn name(&self) -> String {
         self.name.clone()
     }
+    fn state_variable_names(&self) -> Vec<String> {
+        vec![String::from("energy"), String::from("battery_state")]
```
nreinicke, line 49: maybe we should be explicit here and call this `energy_electric` and remove the `energy` value from the default config?

rust/routee-compass-powertrain/src/routee/vehicle/default/ice.rs
```diff
@@ -36,6 +35,9 @@ impl VehicleType for ICE {
     fn name(&self) -> String {
         self.name.clone()
     }
+    fn state_variable_names(&self) -> Vec<String> {
+        vec![String::from("energy")]
```
nreinicke, line 39: Similar to the bev, this could probably become `energy_liquid` and we could pull out `energy` from the config.

rust/routee-compass-core/src/algorithm/search/search_algorithm.rs
```diff
@@ -32,6 +33,7 @@ impl SearchAlgorithm {
         destination: Option<VertexId>,
         graph: Arc<ExecutorReadOnlyLock<Graph>>,
         traversal_model: Arc<dyn TraversalModel>,
+        utility_model: CostModel,
```
nreinicke, line 36: We should probably switch this `cost_model: CostModel`

robfitzgerald:
ok, i'm also seeing big changes, though on my system, your query does complete (after my recent cost fix, just pushed up). here's an energy-optimal route, computed in 40 seconds:

Screen Shot 2023-12-22 at 10 19 59 AM

i ran a distance, time and energy-optimal route via the coefficients using your query:

name | parameters | runtime | tree size | route distance | route time | route energy | total cost |
--- | --- | --- | --- | --- | --- | --- | --- |
distance-optimal | distance = 1 | 2.5 sec | 61044 | 17.60 miles | 42.25 minutes | 4.68kWh | $64.54 |
time-optimal | time = 1 | 5 sec | 131980 | 21.56 miles | 25.14 minutes | 8.11kWh | $54.81 |
energy-optimal | energy = 1 | 40 sec | 186194 | 17.97 miles | 56 minutes | 4.19 kWh | $78.47 |

the trends look correct, and i find them fascinating. runtimes for distance and time look normal. i'm not using your cacheing, that seems it could help here. i want to build routee-compass from the main branch and run this energy-optimal route again, because i'm realizing i don't actually know what to expect for this route's runtimes and results. but, i'm actually quite satisfied from how this stuff looks at-a-glance!

### back to your question

> Do we still have a catch for negative energies?

we have [these lines here](https://github.com/NREL/routee-compass/blob/rjf/cost-abstractions/rust/routee-compass-powertrain/src/routee/vehicle/default/bev.rs#L166) in the BEV VehicleType, is this what you mean? or was there something else that needed to happen further down the line to ensure non-negativity there?

rust/routee-compass-core/src/model/cost/cost_model.rs
```diff
@@ -0,0 +1,234 @@
...
+                Ok((name.clone(), cost))
+            })
+            .collect::<Result<HashMap<String, Cost>, CostError>>()?;
+
+        let total_cost = state_variable_costs
+            .values()
+            .fold(Cost::ZERO, |a, b| a + *b);
+        state_variable_costs.insert(String::from("total_cost"), total_cost);
+
+        let result = serde_json::json!(state_variable_costs);
```
nreinicke, line 189: Would it be possible to add an additional field here, something like `total_realized_cost` that captures the sum of the cost that the model actually optimized for since total cost will be the same regardless of the coefficients that are used.

robfitzgerald:
yes, it appears that the cost estimate is not doing what it should. here's the tree for the energy-optimal route:

Screen Shot 2023-12-22 at 10 47 26 AM

so, basically, Dijkstra's. to investigate why, i think i will need to wait until next week. again, from my comment above, if you recall what the steps were to properly involve VehicleTypes with the EnergyTraversalModel, i would appreciate any hints about where i could be mucking that up.

nreinicke:
Oh interesting, yeah we had a [simple catch ](https://github.com/NREL/routee-compass/blob/0bb79f8faf61373f9aef6b6b29c7a015a3fa62b8/rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs#L102)in the traversal cost method to pin energy to zero but I think your recent commit to cap all costs to the positive domain should fix. I'm okay if we dig into this next week since everything else seems to be working.

Overall this is looking great and I love how much flexibility it gives us. I just finished looking through the changes and have a few comments (hopefully nothing big). But, if any comments are out of scope for finishing today, maybe we just merge and make new issues for anything we don't want to forget over the break?

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
```diff
@@ -96,44 +83,53 @@ impl TraversalModel for EnergyTraversalModel {
...
+            self.service.max_speed,
+            self.service.speeds_table_speed_unit,
+            distance,
+            self.service.output_distance_unit,
+            self.service.output_time_unit.clone(),
+        )?;
+
+        let best_case_result = self
+            .vehicle
+            .best_case_energy_state((distance, self.service.output_distance_unit), state)?;
```
nreinicke, line 129: OH! I think this state just needs to be wrapped in `get_vehicle_state_from_state(state)` to extract the vehicle specific subset of the whole traversal state. 
```suggestion
            .best_case_energy_state((distance, self.service.output_distance_unit), get_vehicle_state_from_state(state))?;
```

rust/routee-compass-core/src/model/cost/vehicle/vehicle_cost_rate.rs
robfitzgerald: i struggled with this name, and like the idea of fixing it, but i'm gonna push this out for future work because the naming convention here also impacts network costs, the cost model, and the query-time argument names.

rust/routee-compass-core/src/model/cost/cost_model.rs
robfitzgerald: that appears as the "cost" field in the base level of the reponse, which gets placed there by the summary output plugin. does that cover your ask?

robfitzgerald: to disambiguate, i just renamed that "route_cost".

also, looking at the summary output plugin, i'm thinking all of this functionality could probably be moved into the CompassApp.apply_output_processing method, and we could remove the JSON summary extensions and summary plugin.

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
robfitzgerald: nice catch! after fixing this, the energy-optimal search runtime is down for me from 40 seconds to 0 seconds. so, now, wondering why that went so fast, and why time and distance are taking so long in comparison. updated table:

name | parameters | runtime | tree size | route distance | route time | route energy | total cost |
--- | --- | --- | --- | --- | --- | --- | --- |
distance-optimal | distance = 1 | 2.5 sec | 61044 | 17.60 miles | 42.25 minutes | 4.68kWh | 64.54 |
time-optimal | time = 1 | 5 sec | 131980 | 21.56 miles | 25.14 minutes | 8.11kWh | 54.81 |
energy-optimal | energy = 1 | 40 sec | 186194 | 19.53 miles | 69 minutes | 2.92 kWh | 92.21 |

the trends all look right here for the distance/time/energy tradeoffs. this is using raw costs so the cost column is kinda nonsense (hence removing the dollar sign).

the a* heuristic is probably still to blame here, given the size of the trees for those searches.

rust/routee-compass-core/src/model/cost/vehicle/vehicle_cost_rate.rs
nreinicke: yeah that makes sense to me

rust/routee-compass-core/src/model/cost/cost_model.rs
nreinicke: > that appears as the "cost" field in the base level of the reponse, which gets placed there by the summary output plugin. does that cover your ask?

ohhh yeah I missed that

> I'm thinking all of this functionality could probably be moved into the CompassApp.apply_output_processing method, and we could remove the JSON summary extensions and summary plugin.

yeah agreed, that would make sense

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
nreinicke: yeah it's also curious why the energy value is down to 2.9 (versus 4 something from the previous version).

something else I noticed for the energy optimal case is that the result["cost_summary"]["cost"]["energy_electric"] = 2.92 but the result["cost"] value is 3.19

perhaps we should just lump this into a new issue and do a deep dive after the break?

nreinicke (approved):
Nice work on adding this all in, it looks great!!

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
robfitzgerald: > result["cost_summary"]["cost"]["energy_electric"] = 2.92 but the result["cost"] value is 3.19

right. i think this would be explained if the state updates were allowed zero/negative-valued energies but the cost function was not.

> perhaps we should just lump this into a new issue and do a deep dive after the break?

i at least have an issue for exploring the poor runtimes (#92) - but please feel free to drop another issue on for exploring the energy results as well.

robfitzgerald:
using our "real-ish" cost factors, i ran a test for distance, energy, time, and "all 3"-optimal routes:

Screen Shot 2023-12-22 at 12 50 48 PM


```
 	name 	 total_cost 	distance 	time 	       energy_electric
0 	distance 21.485037 	17.603244 	42.254257 	4.683044
1 	energy 	 31.347543 	19.529403 	69.753055 	2.924176
2 	time 	 17.560405 	21.555581 	25.139388 	8.111696
3 	all 	 17.515325 	21.009211 	25.796709 	7.276318
```

it all makes sense to me. pretty psyched!

robfitzgerald:
@nreinicke hey, fyi, i was able to get grid search to work with the new setup, here was my solution:

```json
"grid_search": {
    "_ignore": [
      {
        "name": "distance",
        "state_variable_coefficients": {
          "distance": 1,
          "time": 0,
          "energy_electric": 0
        }
      },
      {
        "name": "time",
        "state_variable_coefficients": {
          "distance": 0,
          "time": 1,
          "energy_electric": 0
        }
      },
      {
        "name": "energy",
        "state_variable_coefficients": {
          "distance": 0,
          "time": 0,
          "energy_electric": 1
        }
      },
      {
        "name": "all",
        "state_variable_coefficients": {
          "distance": 1,
          "time": 1,
          "energy_electric": 1
        }
      }
    ]
  }
```

PR #93, opened 2024-01-02
kWh cost factor typo

robfitzgerald:
this itty bitty PR fixes a typo that set the kWh/$ rate at 5 cents instead of 50.


nreinicke (approved):
Yeah I guess it makes sense to set the worst case as the default with the best case being the cost of charging at home (which is probably closer to $0.05 / kwh)

robfitzgerald:
> charging at home 

totally a valid assumption, it's just _not the assumption_ in the comment from the config template:

```rust
# based on $0.50/kWh approximation of DCFC charge rates, $/kWhtype = "factor"
```

PR #97, opened 2024-01-03
prevent grid search recursion

### LINKED ISSUE #96: endless loop when "grid_search" contains "grid_search"
by mistake i injected a value at key "grid_search" which contained a value "grid_search" within it:

```json
 "grid_search": {
    "grid_search": {
      "_details": [ {}, {}, {}, {} ],
     }
  }
```

this causes an infinite loop.

robfitzgerald:
I attempted to inject a grid search but ended up erroneously nesting grid search within itself (see #96). That exposed a logic error that leads to an infinite loop. This fixes that. If a grid search key is found on the input, we do a quick scan of the stringified input for any additional appearance of the substring "grid_search" and produce an error if we find it.

Along the way, I made a simple "debug" input plugin:

```toml
input_plugins = [
  { type = "debug" }
]
```

all it does is pretty prints the current query to the console. this was how i found my error.

Closes #96.


nreinicke (approved):
Good catch and thanks for adding in the new test.

PR #99, opened 2024-01-08
Interpolate Predictor

nreinicke:
Adds a new `InterpolationSpeedGradeModel` that operates by:
 - loading in an existing routee powertrain model 
 - exercising the model over a structured grid of model inputs
 - at predict time, perform a bilinear interpolation to get model outputs

I tested this approach over the training and test set for the data that were used to train the existing random forest models and only observed a link RMSE bump of around 0.00001 relative to the pure random forest model.

The benefit of this model is that it decreases the runtime cost of predicting energy on a link. When testing this over a few Denver metro scenarios (with the addition of some other small performance updates like using `.ok_or_else` instead of `ok_or` to delay string formatting until an error occurs), I'm seeing runtimes improve close to 10x.

### DIFF
--- a/rust/routee-compass-core/src/algorithm/search/search_algorithm_type.rs
+++ b/rust/routee-compass-core/src/algorithm/search/search_algorithm_type.rs
@@ -27,17 +27,14 @@ impl TryFrom<&serde_json::Value> for SearchAlgorithmType {
     /// and returns the SearchAlgorithmType, expected to be a string at the
     /// key "type" on this object.
     fn try_from(config: &serde_json::Value) -> Result<Self, Self::Error> {
-        let type_obj = config
-            .get("type")
-            .ok_or(SearchError::BuildError(String::from(
-                "algorithm config missing 'type' field",
-            )))?;
+        let type_obj = config.get("type").ok_or_else(|| {
+            SearchError::BuildError(String::from("algorithm config missing 'type' field"))
+        })?;
         let alg_string: String = type_obj
             .as_str()
-            .ok_or(SearchError::BuildError(format!(
-                "'type' must be string, found {:?}",
-                type_obj
-            )))?
+            .ok_or_else(|| {
+                SearchError::BuildError(format!("'type' must be string, found {:?}", type_obj))
+            })?
             .into();
         SearchAlgorithmType::from_str(&alg_string)
     }
--- a/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
+++ b/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
@@ -56,8 +56,9 @@ impl PredictionModel for InterpolationSpeedGradeModel {
 impl InterpolationSpeedGradeModel {
     #[allow(clippy::too_many_arguments)]
     pub fn new<P: AsRef<Path>>(
-        model_path: &P,
-        model_type: ModelType,
+        underlying_model_path: &P,
+        underlying_model_type: ModelType,
+        underlying_model_name: String,
         speed_unit: SpeedUnit,
         speed_bounds: (Speed, Speed),
         speed_bins: usize,
@@ -68,9 +69,9 @@ impl InterpolationSpeedGradeModel {
     ) -> Result<Self, TraversalModelError> {
         // load underlying model to build the interpolation grid
         let model = load_prediction_model(
-            "".to_string(),
-            model_path,
-            model_type,
+            underlying_model_name,
+            underlying_model_path,
+            underlying_model_type,
             speed_unit,
             grade_unit,
             energy_rate_unit,
@@ -141,6 +142,7 @@ mod test {
         let model = InterpolationSpeedGradeModel::new(
             &model_path,
             ModelType::Smartcore,
+            "Toyota Camry".to_string(),
             SpeedUnit::MilesPerHour,
             (Speed::new(0.0), Speed::new(100.0)),
             101,
--- a/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
+++ b/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
@@ -1,0 +1,68 @@
+use std::path::Path;
+
+use crate::routee::prediction::prediction_model::PredictionModel;
+use routee_compass_core::{
+    model::traversal::traversal_model_error::TraversalModelError,
+    model::unit::{as_f64::AsF64, EnergyRate, EnergyRateUnit, Grade, GradeUnit, Speed, SpeedUnit},
+};
+use smartcore::{
+    ensemble::random_forest_regressor::RandomForestRegressor, linalg::basic::matrix::DenseMatrix,
+};
+
+pub struct InterpolationSpeedGradeModel {
+    rf: RandomForestRegressor<f64, f64, DenseMatrix<f64>, Vec<f64>>,
+    speed_unit: SpeedUnit,
+    grade_unit: GradeUnit,
+    energy_rate_unit: EnergyRateUnit,
+}
+
+impl PredictionModel for InterpolationSpeedGradeModel {
+    fn predict(
+        &self,
+        speed: (Speed, SpeedUnit),
+        grade: (Grade, GradeUnit),
+    ) -> Result<(EnergyRate, EnergyRateUnit), TraversalModelError> {
+        let (speed, speed_unit) = speed;
+        let (grade, grade_unit) = grade;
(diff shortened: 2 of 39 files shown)


rust/routee-compass-core/src/algorithm/search/search_algorithm_type.rs
```diff
@@ -27,17 +27,14 @@ impl TryFrom<&serde_json::Value> for SearchAlgorithmType {
     /// and returns the SearchAlgorithmType, expected to be a string at the
     /// key "type" on this object.
     fn try_from(config: &serde_json::Value) -> Result<Self, Self::Error> {
-        let type_obj = config
```
robfitzgerald, line 30: thanks for catching these, i need to get into the habit of thinking that ok_or is eager and ok_or_else is lazy!

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -0,0 +1,160 @@
...
+    ) -> Result<Self, TraversalModelError> {
+        // Load random forest binary file
+        let rf_binary = std::fs::read(routee_model_path).map_err(|e| {
+            TraversalModelError::FileReadError(
+                routee_model_path.as_ref().to_path_buf(),
+                e.to_string(),
+            )
+        })?;
+        // Load the random forest regressor so we can exercise it on a linear grid
+        let rf: RandomForestRegressor<f64, f64, DenseMatrix<f64>, Vec<f64>> =
```
robfitzgerald, line 74: noticing that this is smartcore-only. is it possible to add an `underlying: ModelType` to `ModelType::Interpolate`:

```rust
pub enum ModelType {
    Smartcore,
    Onnx,
    Interpolate {
        underlying: ModelType,
        speed_lower_bound: Speed,
        speed_upper_bound: Speed,
        speed_bins: usize,
        grade_lower_bound: Grade,
        grade_upper_bound: Grade,
        grade_bins: usize,
    },
}
```

and then we can load + sample whatever model that is?

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
```diff
@@ -0,0 +1,109 @@
+use ordered_float::OrderedFloat;
+
+pub fn linspace(x0: f64, xend: f64, n: usize) -> Vec<f64> {
```
robfitzgerald, line 3: wondering if you wrote these functions or if you grabbed them from somewhere. they look like something from rosetta code? if so, we should attribute them.

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    if low > 0 && arr[low] >= target {
+        low - 1
+    } else {
+        low
+    }
+}
+
+pub struct BilinearInterp {
```
robfitzgerald, line 33: can this go into a separate file bilinear_interp.rs?

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
```diff
@@ -0,0 +1,109 @@
...
+        let x1 = self.x[x_index + 1].into_inner();
+        let y0 = self.y[y_index].into_inner();
+        let y1 = self.y[y_index + 1].into_inner();
+
+        let q11 = self.values[x_index][y_index];
+        let q12 = self.values[x_index][y_index + 1];
+        let q21 = self.values[x_index + 1][y_index];
+        let q22 = self.values[x_index + 1][y_index + 1];
+
+        let fxy1 = (x1 - x) / (x1 - x0) * q11 + (x - x0) / (x1 - x0) * q21;
```
robfitzgerald, line 75: again, wondering if this is code comes from somewhere else, though i get it if it's just common sense that this is what linear interpolation is.

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+impl InterpolationSpeedGradeModel {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new<P: AsRef<Path>>(
+        routee_model_path: &P,
+        speed_unit: SpeedUnit,
+        speed_bounds: (Speed, Speed),
+        speed_bins: usize,
```
robfitzgerald, line 60: nice, this gives us a way to play with the resolution.

robfitzgerald (changes requested):
this is great. i'm guessing we willl always prefer this approach. i have what i thought was a few comments but i think is maybe a request for changes? let me know what you think.

hello from TRB!

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
nreinicke: yeah that's a great idea, I'll add that in

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
nreinicke: yeah good call, that one was written using an [SO answer](https://stackoverflow.com/a/70840233) as a starting point. I'll add attribution in for it

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
nreinicke: yeah good idea

rust/routee-compass-powertrain/src/routee/prediction/interpolation/utils.rs
nreinicke: Yeah this is based on [one formula for bilinear interpolation](https://en.wikipedia.org/wiki/Bilinear_interpolation). I'll add in the link so we remember where we got it from

nreinicke:
@robfitzgerald - thanks for the review and I hope TRB is going well! I pushed up a couple commits to address your comments whenever you get a chance to re-review. No rush!

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -63,36 +66,41 @@ impl InterpolationSpeedGradeModel {
...
-        let rf: RandomForestRegressor<f64, f64, DenseMatrix<f64>, Vec<f64>> =
-            bincode::deserialize(&rf_binary).map_err(|e| {
-                TraversalModelError::FileReadError(
-                    routee_model_path.as_ref().to_path_buf(),
-                    e.to_string(),
-                )
-            })?;
+        // load underlying model to build the interpolation grid
+        let model = load_prediction_model(
+            "".to_string(),
```
robfitzgerald, line 71: this dummy string feels weird. the reason this is ok is because we actually aren't using this anywhere. it's a nice convention to include a name in the yaml though. something to ponder, whether we can make use of the name or if we should just ignore it.

robfitzgerald (approved):
dig it!

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
nreinicke: yeah I ended up passing the name down into the builder for the interpolation model since it was only up one layer and makes more sense than putting an empty string there.

robfitzgerald:
hey, just reading through again, i have a few questions:
  1. when you checked the new error and found it to be a 0.00001 RMSE diff, was that a spot check, for a single input, or did you run a larger sample across a range of realistic values? would be curious about a plot of that, or gut check, to see there aren't any big outliers or parts where the quantized result creates bigger distortions
  2. does this satisfy us wrt runtimes? open question. if i have a little time i'll try running a grid search test to see what it's doing to the different trip objectives and corresponding search runtimes

nreinicke:
> was that a spot check, for a single input, or did you run a larger sample across a range of realistic values?

To get that number I applied both models to a hold out test set of about 600k links and then looked at the difference in their respective RMSE values. Here are the actual RMSE values over the test set:

__Interpolation RMSE__:       0.00121274
__Random Forest RMSE__:   0.00121156

Here's a scatter plot of the predictions of each model:

image

They look really close on the small links and the interpolation model might slightly overestimate energy when the energy values get larger. 

> does this satisfy us wrt runtimes? 

yeah good question, maybe we can run some more spot checks and then try for a national run to see if runtime is a bottleneck or not?

PR #102, opened 2024-01-12
Version 0.6.1

nreinicke:
After running into [yet another set of build errors](https://github.com/conda-forge/nrel.routee.compass-feedstock/pull/7/checks?check_run_id=20433194404) with the onnx runtime, I'm proposing that we omit it from our default build for the time being since we're not actually using it. We can come back to this and re-visit if we decide that it's valuable to have.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -47,7 +47,7 @@ name = "2012_Ford_Focus"
 type = "single_fuel"
 # the file for the routee-powertrain model
 model_input_file = "models/2012_Ford_Focus.bin"
-# what underlying machine learn framework to use [smartcore | interpolate]
+# what underlying machine learn framework to use [smartcore | interpolate | onnx]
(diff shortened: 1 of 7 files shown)


docs/config.md
```diff
@@ -47,7 +47,7 @@ name = "2012_Ford_Focus"
 type = "single_fuel"
 # the file for the routee-powertrain model
 model_input_file = "models/2012_Ford_Focus.bin"
-# what underlying machine learn framework to use [smartcore | onnx]
```
robfitzgerald, line 50: noting this comment, maybe it should still list onnx as a possibility

robfitzgerald (approved):
totally agree!

docs/config.md
nreinicke: yeah good point, I'll add that back in

PR #103, opened 2024-01-13
Turn Access Cost

nreinicke:
Implements a turn access cost into the energy traversal model by loading in a table of edge headings and then computing the difference between headings at runtime to get a turn angle. This turn angle gets translated into a `Turn` enum and appropriate turn costs are added. I started by taking turn costs [similar to Valhalla](https://github.com/valhalla/valhalla/blob/5540cd7d0fcf67b5a1b8ee16fc5d6cbd2360b4c9/src/sif/autocost.cc#L40) but we can discuss if that makes sense or not.


rust/routee-compass-core/src/model/road_network/turn.rs
```diff
@@ -0,0 +1,32 @@
...
+pub enum Turn {
+    NoTurn,
+    SlightRight,
+    SlightLeft,
+    Right,
+    Left,
+    SharpRight,
+    SharpLeft,
+    UTurn,
+}
```
robfitzgerald, line 12: seems we could make Turn deserializable and then expose an optional `HashMap<Turn, Time>` mapping to some turn delays configuration in the future to remove the hard-coded values in the energy traversal model

rust/routee-compass-core/src/model/unit/time_unit.rs
```diff
@@ -13,7 +13,7 @@ pub enum TimeUnit {
 }
 
 impl TimeUnit {
-    pub fn convert(&self, value: Time, target: TimeUnit) -> Time {
+    pub fn convert(&self, value: Time, target: &TimeUnit) -> Time {
```
robfitzgerald, line 16: right, good catch here

rust/routee-compass-core/src/model/road_network/edge_heading.rs
```diff
@@ -0,0 +1,68 @@
+use serde::Deserialize;
+
+#[derive(Copy, Clone, Deserialize)]
+pub struct EdgeHeading {
+    start_heading: i16,
+    end_heading: Option<i16>,
```
robfitzgerald, line 6: future work we could make Headings a Unit type in RouteE Compass

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
```diff
@@ -90,13 +91,58 @@ impl TraversalModel for EnergyTraversalModel {
...
+        src: &Edge,
         _v2: &Vertex,
-        _dst: &Edge,
+        dst: &Edge,
         _v3: &Vertex,
-        _state: &TraversalState,
+        state: &TraversalState,
     ) -> Result<Option<TraversalState>, TraversalModelError> {
-        Ok(None)
+        match self.service.headings_table.as_deref() {
```
robfitzgerald, line 100: this feels like a behavior that should be shared across traversal models, we'll have to think about the right abstraction there before we start adding any more of these Edge->Edge update features.

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
```diff
@@ -90,13 +91,58 @@ impl TraversalModel for EnergyTraversalModel {
...
+                        // 3.5 second penalty for sharp left
+                        Time::new(3.5)
+                    }
+                    Turn::UTurn => {
+                        // 9.5 second penalty for U-turn
+                        Time::new(9.5)
+                    }
+                };
+                let time = TimeUnit::Seconds.convert(time_cost, &self.service.output_time_unit);
+                let updated_state = add_time_to_state(state, time);
```
robfitzgerald, line 142: there should really be an energy update here too. are the Powertrain models well behaved for speed and grade of 0, to give us a kind of "idling energy", or is that something we need to supplant with a configuration value? do we have those values?

robfitzgerald (commented):
heya, interesting work. thanks for throwing in the few quality of life improvements too.

i'm pretty sure we can just roll this in, but wanted to ask a few questions first before approving. mostly, this PR has me wondering about how we can share this kind of feature across traversal models without having to hard-code it every time, like if we had a good abstraction that could be referenced completely via configuration. but maybe after a quick back and forth we can accept this while including a few new issues for establishing the pattern for making these kinds of improvements in a more permanent way.

rust/routee-compass-core/src/model/road_network/turn.rs
nreinicke: Yeah it would be nice to expose the turn cost to the configuration. I struggled to land on a way to do this that felt right and so maybe we could have a design discussion about ways to abstract this.

rust/routee-compass-core/src/model/road_network/edge_heading.rs
nreinicke: Yeah that's a good idea

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
nreinicke: Yeah agreed.

rust/routee-compass-powertrain/src/routee/energy_traversal_model.rs
nreinicke: Right, I think this merits it's own exploration to decide on the best way to represent energy turn costs. It sounds like Jake and Mingdong are planning to take a pass at exploring this.

Also, following up on our discussion earlier today, the powertrain models are not well defined for speeds of 0 and so we would need to provide an explicit "idling energy" if we want to represent the model in those situations. I think we have (or could easily get) the idle costs for each vehicle

robfitzgerald (approved):
cool thanks for answering my questions. i can help with some of the future tasks i described, would you mind making issues for them?

PR #108, opened 2024-02-02
Ndr/rtree mem fix

nreinicke:
When running some simple performance tests I noticed that the app was failing when loading the edge rtree plugin. This is just a simple fix to not clone the geometry and road class strings when building the plugin but rather pass ownership directly to the rtree.


robfitzgerald (approved):
great find!

PR #110, opened 2024-02-02
Implement Allocative

nreinicke:
I just wanted to share this proof of concept for keeping an eye one memory allocation during runtime. The [allocative](https://docs.rs/allocative/latest/allocative/index.html) crate lets you implement the `Allocative` trait on your internal structs and then will give you a dynamic size at runtime. I slapped together a quick demo of this by implementing it for the graph struct (using some hacky newtypes to get it running) but when running the denver scenario it prints out a nice line:

```console
[2024-02-02T22:46:58Z INFO  routee_compass::app::compass::compass_app] graph size: 0.112550424 GB
```

It also lets us look deeper by writing a flamegraph. For example if I use [flamegraph.pl](https://github.com/brendangregg/FlameGraph/blob/master/flamegraph.pl):

```bash
./flamegraph.pl flamegraph.out > flamegraph.csv
```

I get a nice look at the proportion of memory:

image

I didn't have the time to go further with this but wanted to share and get feedback in case it's something that we might want to look at in the future.


nreinicke:
yeah that sounds good to me!

nreinicke:
@robfitzgerald - this should be ready for your review now. The bulk of this work was just to implement the `Allocative` trait across many of our internal data structures, sometimes having to introduce a new type to get the trait implemented.

rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
```diff
@@ -188,6 +194,25 @@ pub fn run_a_star(
...
+        let output = flamegraph.finish_and_write_flame_graph();
+
+        let search_name = match target {
+            None => format!("{}_to_all", source),
+            Some(tid) => format!("{}_to_{}", source, tid),
+        };
+
+        let mut flamegraph_file =
+            File::create(format!("search_memory_flamegraph_{}.out", search_name)).unwrap();
+        flamegraph_file.write_all(output.as_bytes()).unwrap();
```
robfitzgerald, line 213: 1. could we pass debug_mode as a parameter to the run_a_star method(s), and perform this env::var("_") check out in the SearchApp scope, so it only gets checked once
  2. could we set the filename based on the query source + target VertexIds so that if multiple queries are run, they all write flamegraphs to different destinations? or would that be silly?

robfitzgerald (changes requested):
looks good. i've been trying to read up to convince myself that these newtype wrappers are in fact zero-cost, and i'm still not sure. there are [discussions](https://www.reddit.com/r/rust/comments/p0ul6b/when_zero_cost_abstractions_arent_zero_cost/) about the conditions that make newtypes zero-cost. maybe we both keep an eye on the allocations of these wrapper types while continuing to profile Compass.

i've got a small request for a change but all in all, great to see this rolled in, thanks for plumbing this into the app!

rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
nreinicke: > could we pass debug_mode as a parameter to the run_a_star method(s), and perform this env::var("_") check out in the SearchApp scope, so it only gets checked once 

yeah that's smart; I also stumbled across [this SO answer](https://stackoverflow.com/a/39205417) about just doing conditional compilation for this section so we don't even include it in a release build.

> could we set the filename based on the query source + target VertexIds so that if multiple queries are run, they all write flamegraphs to different destinations? or would that be silly?

yeah I was wondering about this; I think it's setup now to write a file for each search by using the `search_name` variable which is a combo of the source and target ids.

robfitzgerald: ya! debug_assertions! cool, let's do that!

rust/routee-compass/src/app/search/search_app.rs
```diff
@@ -122,8 +122,8 @@ impl SearchApp {
                 );
                 Ok(SearchAppResult {
                     route,
-                    tree,
-                    search_start_time,
+                    tree: Some(tree),
```
robfitzgerald, line 125: did we want to unravel this Option wrap per [this discussion](https://github.com/NREL/routee-compass/issues/114#issuecomment-1933067350)?

robfitzgerald (changes requested):
noticed also this other thing

robfitzgerald (approved):
:beers:

PR #111, opened 2024-02-05
Ndr/cost model performance

nreinicke:
Applies some minor modifications to the cost model to use iterators instead of allocating a vector for the cost model aggregation.

Also flattens the state variable coefficients, vehicle rates and network rates into a fixed size vector for index based lookup.

### DIFF
--- a/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
+++ b/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
@@ -465,8 +465,7 @@ mod tests {
                     )])),
                     Arc::new(HashMap::new()),
                     CostAggregation::Sum,
-                )
-                .unwrap();
+                )?;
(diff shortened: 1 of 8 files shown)


nreinicke:
> did this change make a dent in your profiles?

it did reduce the search runtime a bit but nothing to write home about. I'm hoping that there are some other areas we haven't touched that might result in even bigger gains..

nreinicke:
> did this change make a dent in your profiles?

Actually, when I tested this out on a national run it had more of an effect. Before these changes I got these numbers for running batches of long queries over the national graph:

```console
average route runtime: 137.328 seconds
average seconds per distance: 0.10788 seconds
average milliseconds per tree edge: 0.02720 milliseconds
```

And after

```console
average route runtime: 33.852 seconds
average seconds per distance: 0.02219 seconds
average milliseconds per tree edge: 0.00545 milliseconds
```

robfitzgerald:
awesome! great stuff nick 👨‍🎤 🎸

PR #113, opened 2024-02-06
a prototype of creating AccessModels

### LINKED ISSUE #104: [Discussion] Turn Cost Abstracion
In #103 we introduced a static time turn cost to the energy traversal model but acknowledged that this functionality could be useful across different traversal models. In order to share the logic we would need to come up with a clean abstraction for how to represent a turn (mapping turn degrees to turn type) and how much cost to apply for said turn. We might also consider the idea of using road class transitions as an indicator of time or energy penalty as a motivation for how to implement this abstraction.

### LINKED ISSUE #109: abstraction around access modeling, starting with turn costs
our turn cost model is hard-wired to the energy model, but should be accessible from other traversal models. 

in general, we may want many such access models under the hood of a traversal model:

```rust
struct MyTraversalModel {
  access_models: Vec<AccessModel>
}
```

that we can step through and apply at each step of a search.

with this abstraction, we can re-write turn delays as its own access model which can be built from configuration.

robfitzgerald:
this PR makes AccessModel a top-level Model module (routee-compass-core::model::access). the access_edge method has been removed from traversal models. a few default implementations are in place:
  - no_access_model: a simple noop model
  - combined: more than one access model
  - turn_delay: assigns turn delay times from a table based on the heading of the transition

an example configuration for a turn_delay access model:

```toml
[access]
type = "turn_delay"
edge_heading_input_file = "../data/tomtom_denver/edges-headings-enumerated.csv.gz"
[access.turn_delay_model]
type = "tabular_discrete"
time_unit = "seconds"
[access.turn_delay_model.table]
no_turn = 0.0
slight_right = 0.5
right = 1.0
sharp_right = 1.5
slight_left = 1.0
left = 2.5
sharp_left = 3.5
u_turn = 9.5
```

wanted to leave a variation of turn delays that incorporates road classes but decided to let that sit and left a comment as a placeholder for it.

along the way, i fixed some missing feature injections. SpeedTraversalModel::state_features() wasn't returning it's features and EnergyTraversalModel::state_features() wasn't returning the features of it's underlying time model (touches #145).

i've completed this implementation but i'm unsure if the results are correct, and it may be connected with the discussion around how we pass around units. that said, i did see that my output.json's geometry properties did include access costs, which implies the logic was being called. leaving follow-up to our next dev check-in.

Closes #104.
Closes #109.


nreinicke:
This looks great. I just tested out on the osm data and without a turn delay model I get a time of 10.6 minutes and with a turn delay model I get a time of 10.9 seconds so it seems like it's being incorporated!

I did have to change [the dataset script ](https://github.com/NREL/routee-compass/blob/f049e14030b75bddd8316f4536f42c7acda4d902/python/nrel/routee/compass/io/generate_dataset.py#L158)to use `arrival_heading` and `departure_heading` and so we should probably fix that really quick (I can push a commit here if you want). Other than that it looks good to me!

robfitzgerald:
You sure you pulled down the latest headings dataset via lfs?

nreinicke:
> You sure you pulled down the latest headings dataset via lfs?

I was actually testing on the OSM dataset that gets built from our [`generate_dataset`](https://github.com/NREL/routee-compass/blob/main/python/nrel/routee/compass/io/generate_dataset.py) script. I think we just need to update how we're writing our headings file to match the new format (previously I used `start_heading` and `end_heading` but now the rust code expects `arrival_heading` and `destination_heading`).

robfitzgerald:
> previously I used start_heading and end_heading but now the rust code expects arrival_heading and destination_heading

right. this was a conscious choice at some point. but, it also is a choice that was made within the fever dream of working on routee while you were on PTO. it wasn't _necessary_, and now i'm having trouble convincing myself that this was _helpful_. but i recall i was confused by the meaning of "start" and "end" too. maybe there's a better choice? but i'll update the osm generation code for now so it works!

nreinicke (approved):
Yeah I think it's fine to use the `arrival` `destination` distinction, thanks for pushing up that fix!

PR #120, opened 2024-02-12
Rjf/cost model invariants

### LINKED ISSUE #119: cost model refactor: index out of bounds error
### description

while running a custom RouteE scenario (the MEP integration), i got this panic from an index out of bounds issue. the error points to the CostModel code changes that moved from a HashMap to a Vec representation and perhaps didn't preserve some invariant from the original code. i was running multiple threads so the error message looks funny; cost_model.rs line 40 or 38 seem the target, and indeed line 38 does involve indexing into a vector:

```rust
state_variable_coefficients[*idx] = *coef;
```

### error message

```
thread 'thread '<unnamed>' panicked at <unnamed>/Users/rfitzger/dev/nrel/routee/routee-compass/rust/routee-compass-core/src/model/cost/cost_model.rs' panicked at /Users/rfitzger/dev/nrel/routee/routee-compass/rust/routee-compass-core/src/model/cost/cost_model.rs:thread '38:<unnamed>:38' panicked at 40:/Users/rfitzger/dev/nrel/routee/routee-compass/rust/routee-compass-core/src/model/cost/cost_model.rs:
40:index out of bounds: the len is 2 but the index is 2:
38
index out of bounds: the len is 2 but the index is 2:
40:
index out of bounds: the len is 2 but the index is 2
```

robfitzgerald:
this PR fixes some logic errors in the cost model.

##### state indices vs cost model indices

we're looking a lot of stuff by index now which gets ambiguous in the cost model because there are _two kinds of indices_ at play there. the thing i'm calling a _state index_ is the index of a particular state variable in a state vector. the other, maybe worth calling a _model index_ is a position in one of the vectors stored on the CostModel instance, such as the vehicle rate parameter vector. it turns out these were mixed up.

##### empty defaults

i got confused a bunch by unexpected behavior surrounding cost model defaults. i basically ripped out any defaults to make the user more responsible for properly set the cost model values and so they won't waste an hour looking around in the repo for an explanation of why certain things were happening. i _think_ this doesn't mess with any of our capabilities we wanted.

##### reporting runtimes

i made some changes to the runtime reporting. i wanted to focus on output processing, and so now there's a runtime for creating the tree and route outputs. i wanted also to have one for the output plugin processing, but i stopped short of it since the to_disk plugin dumps an output row before we get to the outer scope of running the output plugins. future work to set that up.

Closes #119.

### DIFF
--- a/python/nrel/routee/compass/resources/osm_default_distance.toml
+++ b/python/nrel/routee/compass/resources/osm_default_distance.toml
@@ -14,6 +14,8 @@ distance_unit = "miles"
 [cost.vehicle_state_variable_rates.distance]
 type = "factor"
 factor = 0.655
+[cost.state_variable_coefficients]
+distance = 1
 
 [plugin]
 input_plugins = [
--- a/python/nrel/routee/compass/resources/osm_default_energy.toml
+++ b/python/nrel/routee/compass/resources/osm_default_energy.toml
@@ -37,6 +37,12 @@ factor = 3.120
 type = "factor"
 factor = 0.50
 
+[cost.state_variable_coefficients]
+distance = 1
+time = 1
+energy_liquid = 1
+energy_electric = 1
+
 [traversal]
 type = "energy_model"
 speed_table_input_file = "edges-posted-speed-enumerated.txt.gz"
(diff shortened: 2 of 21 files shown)


nreinicke (approved):
Yep, this makes sense to decouple the 'state_index' from the 'model_index' since state index could be anything and be in any order whereas we want to keep track of them in a fixed order. Thanks for catching this and fixing it!

PR #121, opened 2024-02-14
re-use solution state for search

nreinicke:
This is a simple modification to the a star algorithm code such that we don't store two copies of the traversal state during the search. Now, we just grab the current state from the solution. 

Before making this change, the memory footprint of the algorithm (at the end of the search) looked like this:

image

After this change, the memory footprint looks like:

image


nreinicke:
Actually... I think the whole frontier data structure might be redundant. I attempted to pull that out and just rely on the solution for pulling all the information that the frontier object was providing during the search.

robfitzgerald (approved):
nice catch, i think you're right, the frontier collection was an artifact from duplicating pseudocode but ends up duplicating some aspects of the solution collection. that should be a big help with long-distance trips. :guitar:

nreinicke:
I did do a simple test by running the code before and after this change over a set of sample queries in the denver metro area and am finding that the summaries, routes and trees are exactly the same in both cases.

PR #122, opened 2024-02-14
Rjf/output json mutable

### LINKED ISSUE #116: Output Plugins should modify a mutable JSON result
when an output plugin is run, it makes a copy of the upstream result JSON, performs work, and returns it wrapped in an array, with possibly other JSON objects. for very large inputs, this can lead to a lot of unnecessary allocations.

to avoid this, we can modify the output plugin signature so that the plugin modifies a `mutable` JSON and returns nothing:

```rust
pub trait OutputPlugin: Send + Sync {
    fn process(
        &self,
        output: &mut serde_json::Value,
        result: &Result<SearchAppResult, CompassAppError>,
    ) -> Result<(), PluginError>;
}
```

robfitzgerald:
This PR attempts to improve Compass performance by making the OutputPlugin process method a mutable update on the output JSON.

the existing code requires a copy for any processing, and wraps that copy in a vector to allow flat_map semantics (producing multiple output rows from the upstream JSON). for very large batches, this creates a lot of unnecessary memory pressure which grows over the course of a run.

### remaining work

one downside immediately is that this nullifies our `to_disk` plugin, since that plugin would "void" a response after writing it. my working plan is to move this logic permanently into the CompassApp and activate the logic based on the presence of an output file location specified either by an argument on the run method or by some default specified at the base level of the TOML. thoughts? 

Closes #116.

### DIFF
--- a/rust/routee-compass/src/app/compass/response/response_writer.rs
+++ b/rust/routee-compass/src/app/compass/response/response_writer.rs
@@ -22,7 +22,7 @@ impl ResponseWriter {
                 filename: _,
                 file,
                 format,
-                delimiter,
+                delimiter: _,
             } => {
                 let file_ref = Arc::clone(file);
                 let mut file_attained = file_ref.lock().map_err(|e| {
@@ -53,7 +53,7 @@ impl ResponseWriter {
                 filename,
                 file,
                 format,
-                delimiter,
+                delimiter: _,
             } => {
                 let file_ref = Arc::clone(file);
                 let mut file_attained = file_ref.lock().map_err(|e| {
--- a/rust/routee-compass/src/app/compass/response/response_writer.rs
+++ b/rust/routee-compass/src/app/compass/response/response_writer.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 pub enum ResponseWriter {
+    NoOutputWriter,
     RunBatchFileSink {
         filename: String,
         file: Arc<Mutex<File>>,
@@ -18,6 +19,7 @@ pub enum ResponseWriter {
 impl ResponseWriter {
     pub fn write_response(&self, response: &serde_json::Value) -> Result<(), CompassAppError> {
         match self {
+            ResponseWriter::NoOutputWriter => Ok(()),
             ResponseWriter::RunBatchFileSink {
                 filename: _,
                 file,
@@ -49,6 +51,7 @@ impl ResponseWriter {
 
     pub fn close(&self) -> Result<String, CompassAppError> {
         match self {
+            ResponseWriter::NoOutputWriter => Ok(String::from("")),
             ResponseWriter::RunBatchFileSink {
                 filename,
                 file,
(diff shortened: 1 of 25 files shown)


nreinicke:
Nice, this is looking good. Yeah I like the idea of refreshing how we write to a file. I'm not totally sure about the optimal way to do this but it could make sense to consider different use cases like:
 - running a single batch of queries and writing all the results to one file
 - running multiple batches of queries and writing the results to one file
 - running multiple batches of queries and writing the results to multiple files
 - running one or many batches of queries and just keeping the results in memory to do _something_ with them later

robfitzgerald:
your points are great. i'm going to roll in some work to get us moving in the direction you are describing. i think it boils down to two things:
  - an app-level policy for memory retention
  - an app-level policy for file writing

i'm going to implement these, but in particular, only a basic output policy that replicates our json newline output format.

robfitzgerald:
hmm. actually, maybe i'll revise my earlier statement:

 - an application configuration-level policy for memory retention
 - a `run`-level policy for file writing

and, maybe that means that this latter policy is defined when someone executes the run command by passing an optional configuration object:

```rust
pub fn run(
        &self,
        queries: Vec<serde_json::Value>,
        config: Option<serde_json::Value>
    ) -> Result<Vec<serde_json::Value>, CompassAppError> 
```
🤷‍♂️

robfitzgerald:
@nreinicke this is now ready for review.

the default configuration now starts with this top section which sets "no file writing" as the default but can be swapped out for a version that discards output from RAM while writing to disk:

```toml
parallelism = 2
search_orientation = "vertex"
response_persistence_policy = "persist_response_in_memory"

# # example file output policy
# response_persistence_policy = "discard_response_from_memory"
# [response_output_policy]
# type = "run_batch_file_sink"
# filename = "output.json"
# format = { type = "json", newline_delimited = true }

[response_output_policy]
type = "no_output"

```

rust/routee-compass/src/app/compass/response/response_writer.rs
```diff
@@ -0,0 +1,78 @@
...
+            } => {
+                let file_ref = Arc::clone(file);
+                let mut file_attained = file_ref.lock().map_err(|e| {
+                    CompassAppError::ReadOnlyPoisonError(format!(
+                        "Could not aquire lock on output file: {}",
+                        e
+                    ))
+                })?;
+
+                // if write_delimiter {
```
robfitzgerald, line 37: noting, this solution doesn't really give us the ability to write a delimiter, since the write can be called from different threads and we cannot know when we are between rows. to get there, i would want to wrap the writer itself in a mutex, hold a buffer, and then apply a delimiter when appropriate. but since reproducing our current capability was the goal, this is what we've got.

nreinicke (approved):
This makes sense to me, I like that we can pass in a config at the run method to set any of these variables. Nice work!

PR #125, opened 2024-02-16
Rjf/input plugins mutable

### LINKED ISSUE #124: Input Plugins should modify a mutable JSON result
Let's do #116 but also for input plugins.

```rust
pub trait InputPlugin: Send + Sync {
    fn process(&self, input: &mut serde_json::Value) -> Result<(), PluginError>;
}
```

We still want to support grid search here. Perhaps an idiomatic way to allow this is that the plugin can replace the input with a `serde_json::Value::Array`. the outer scope will do a quick match after calling an input plugin, and if the value has been turned into an array, it will unpack and call subsequent plugins on all objects within the array.

robfitzgerald:
this PR modifies the signature of input plugin so that it relies on mutable semantics, like the recent changes to output plugins, in order to reduce allocations and improve runtime performance, especially for very large input sets.

Closes #124.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -21,9 +21,8 @@ use crate::{
         search::{search_app::SearchApp, search_app_result::SearchAppResult},
     },
     plugin::{
-        input::input_plugin::InputPlugin,
-        output::{initial_output_ops as out_ops, output_plugin::OutputPlugin},
-        plugin_error::PluginError,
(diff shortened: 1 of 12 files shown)


nreinicke (approved):
This looks good to me and I appreciate the detail error messages for invalid plugins.

robfitzgerald:
> I appreciate the detail error messages for invalid plugins

@nreinicke a note on that. because this is happening via mutable updates, there are times where the query cannot be logged because of the borrow checker. i think that there's no way around it (without creating a copy _before_ the processing, which counter-acts the whole intention here), but you may want to review in case you can think through a way to log those mutably-borrowed query objects.

PR #126, opened 2024-02-16
use indoc for error formatting

nreinicke:
A couple of minor updates:
 - use indoc to format the multiline error message a little nicer
 - pass the query as a mutable reference to the `package_error` function


robfitzgerald:
just noting that this didn't pass the python test actions, looks like just a formatting issue

robfitzgerald (approved):
nice, that's a helpful addition, looks good!

PR #127, opened 2024-02-16
bug fixes

robfitzgerald:
caught a few bugs after my recent PR:
- forgot to switch to mutable update for edge rtree plugin
- forgot to make flattening JSON an operation between each input plugin, so grid search result would stay nested


nreinicke (approved):
Yep, this makes sense to me!

PR #130, opened 2024-02-23
New wrapper macro

nreinicke:
Adds a new `#[pybindings]` macro to wrap any struct that implements the new `CompassAppBindings` trait with the appropriate python methods. This allows a sub crate to implement all of the same bindings as the original crate but provide custom functionality. For example, a crate that imports routee-compass could have its own python application with custom modules injected into the builder

```rust
#[pybindings]
pub struct CustomAppPy {
    app: CompassApp,
}

impl CompassAppBindings for CustomAppPy {
    fn from_config_toml_string(
        config_string: String,
        original_file_path: String,
    ) -> Result<Self, CompassAppError>
    where
        Self: Sized,
    {
        let mut builder = CompassAppBuilder::default();

        // inject custom builders here

        let app =
            CompassApp::try_from_config_toml_string(config_string, original_file_path, &builder)?;
        Ok(CustomAppPy { app })
    }
    fn app(&self) -> &CompassApp {
        &self.app
    }
}
```


robfitzgerald:
dude. slick.

i've only just skimmed, gonna load this now and read it in the IDE, but, i think you nailed it.

python/nrel/routee/compass/compass_app.py
```diff
@@ -73,7 +77,7 @@ def from_dict(cls, config: Dict, working_dir: Optional[Path] = None) -> CompassA
         """
         path_str = str(working_dir.absolute()) if working_dir is not None else ""
         toml_string = toml.dumps(config)
-        app = CompassAppWrapper._from_config_toml_string(toml_string, path_str)
+        app = cls.get_constructor()._from_config_toml_string(toml_string, path_str)
```
robfitzgerald, line 80: we don't need get_constructor, that's what `cls` is, you could just call `cls._from_config_toml_string` directly.

rust/routee-compass-py/src/app_wrapper.rs
```diff
@@ -1,114 +1,31 @@
...
-    }
-
-    pub fn graph_get_in_edge_ids(&self, vertex_id: usize) -> PyResult<Vec<usize>> {
-        ops::get_in_edge_ids(self, vertex_id)
-    }
-
-    #[classmethod]
-    pub fn _from_config_toml_string(
-        _cls: &PyType,
+impl CompassAppBindings for CompassAppWrapper {
```
robfitzgerald, line 15: smooth

rust/routee-compass-macros/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+extern crate syn;
+
+use proc_macro::TokenStream;
+use proc_macro_error::proc_macro_error;
+use quote::quote;
+use syn::{parse_macro_input, DeriveInput};
+
+#[proc_macro_error]
+#[proc_macro_attribute]
+pub fn pybindings(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
robfitzgerald, line 13: fn comments

rust/routee-compass/src/app/bindings.rs
```diff
@@ -0,0 +1,109 @@
...
+        road_network::{edge_id::EdgeId, vertex_id::VertexId},
+        unit::{as_f64::AsF64, DistanceUnit},
+    },
+};
+
+use super::compass::{compass_app::CompassApp, compass_app_error::CompassAppError};
+
+use crate::app::search::search_app_graph_ops::SearchAppGraphOps;
+
+pub trait CompassAppBindings {
```
robfitzgerald, line 15: comments

rust/routee-compass/src/app/compass/compass_app.rs
```diff
@@ -59,6 +59,22 @@ pub struct CompassApp {
     pub response_output_policy: ResponseOutputPolicy,
 }
 
+impl CompassApp {
+    pub fn try_from_config_toml_string(
```
robfitzgerald, line 63: comments

rust/routee-compass-macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "routee-compass-macros"
+version = "0.1.0"
```
robfitzgerald, line 3: should this match the versions of the rest of the routee crates?

robfitzgerald (changes requested):
dude, this ends up being a great use of macros to solve our python API problem. i have a small change request on the python code but the main ask i have is for comments on the new stuff. i would like to have some info describe how the pieces fit together. like, why is "bindings.rs" and the macros "lib.rs" file separate/how do they fit together? i think we could also use a README.md in the new crate describing how people can use the macro for their own python app.

but man i'm excited to use this!

python/nrel/routee/compass/compass_app.py
robfitzgerald: or, am i getting this wrong, because it's coming from a macro?

robfitzgerald:
after reading through the tomtom repo i'm seeing how get_constructor is used, to disambiguate the correct constructor through inheritance. as my question and self-answer shows, it would be helpful to add a docstring comment to get_constructor to make clear what's going on there.

nreinicke:
That all sounds good, I'll take a pass through and add in some more comments to try to make it clear how to use this.

rust/routee-compass-macros/Cargo.toml
nreinicke: yeah that probably makes the most sense

robfitzgerald (approved):
rockin

PR #131, opened 2024-02-26
Update osm example notebook

nreinicke:
Updates the example notebook to include recent updates like vehicle state variable rates and state variable coefficients. Also updates to use Denver as the example map in an attempt to find an interesting route that showcases tradeoffs between optimizing for time, energy or cost.


robfitzgerald:
i needed to also update the notebook in order to test all of my changes, incl. config TOML updates. i accept the blame here, will be happy to deal with the conflicts and propose a blend of our notebook updates after i merge in the energy model updates.

nreinicke:
superseded by #156

PR #136, opened 2024-03-08
Rjf/state model

### LINKED ISSUE #115: Abstraction for TraversalState
A TraversalModel is currently responsible for knowing what the meaning of a TraversalState and providing methods for interacting with the state. many methods may create a new state vector in order to do update work or otherwise. in a few ways, there is actually a need to share the interface for a TraversalState between different modules:
- TraversalModel: module that does state updates due to link traversals
  - EnergyTraversalModel has specific state for the Vehicle model
- AccessModel: forthcoming module that does state updates dues to link access rules
- FrontierModel: module that uses state to determine if links are valid for a traversal

In order to share the knowledge of the state between all of the above, we could create an abstraction that provides methods for interacting with a state vector and share it between the above modules. a traversal state object would still be a Vec<StateVar> but the `TraversalStateModel` would track name to index information, would be the factory for cloning/allocating a new mutable vector, and perhaps even hold unit information.

this solves a performance issue by allowing the downstream models to operate on a mutable state vector created by the TraversalStateModel and share the same behaviors for interacting with that vector during a search iteration. this way, a new vector is created once per iteration. it also allows us to automate vector sizing for any TraversalState to the known state representation size using `Vec::with_capacity(n)` to ensure all state vectors are as small as possible.

### LINKED ISSUE #134: replace RwLock with Arc
early in developing RouteE Compass, while digging around in stack overflow for examples of shared thread-safe read-only data, i settled on creating a wrapper class for RwLock that does not expose the write method to threads that use it. the thinking is described in the doc comment at the top of read_only_lock.rs:

```rust
/// there are no read-only locks in the existing rust library
/// that prevent executors from having write access. this pair
/// of classes allows a driver application to instantiate a RwLock
/// on an object, and then issue read-only clones of that lock to
/// executors in a parallel setup.
///
/// see https://stackoverflow.com/a/68908523
```

about 6 months of learning Rust, and it's become obvious that the feature i was looking for is provided by `Arc` all along. from the rust library:

```rust
/// A thread-safe reference-counting pointer. 'Arc' stands for 'Atomically
/// Reference Counted'.
///
/// ...
///
/// Shared references in Rust disallow mutation by default, and `Arc` is no
/// exception: you cannot generally obtain a mutable reference to something
/// inside an `Arc`. If you need to mutate through an `Arc`, use
/// [`Mutex`][mutex], [`RwLock`][rwlock], or one of the [`Atomic`][atomic]
/// types.
```

we can remove the RwLock wrapper and simply work directly with `Arc` in `SearchApp`.

robfitzgerald:
This PR adds a `StateModel` to Compass which manages the creation and inspection of search state vectors. along the way, 
- traversal model updates were transitioned to mutable updates
- removed RwLock wrappers in SearchApp
- some changes to how summaries look

Closes #134.
Closes #115.


robfitzgerald:
@nreinicke I am with Sarah for a baby appt., but just wanted to message that I think I omitted unit conversion handling. I forget. I'll try and review it when we get home.

nreinicke:
> just wanted to message that I think I omitted unit conversion handling. I forget. I'll try and review it when we get home.

That sounds good. 

I was also wondering if you have an example of a config that is working for this update, when I tried to load one of the tomtom denver configs I get:

```
[2024-03-08T21:42:54Z ERROR routee_compass] Could not build CompassApp from config file: expected field state for TOML not found
Error: CompassConfigurationError(ExpectedFieldForComponent("state", "TOML"))
```

I think I just need to add a section to configure the state but not exactly sure what that looks like. l'll also look in the code for more details but thought you might have a ready example.

robfitzgerald:
Right, I plan to put together a PR on the TomTom side of things with these
updates. But you can see an example in the speed_test.toml file in the
routee_compass crate that would work for a speed traversal model.

rust/routee-compass-core/src/model/state/state_model.rs
```diff
@@ -0,0 +1,358 @@
+use super::{
+    state_error::StateError, state_feature::StateFeature, state_model_entry::StateModelEntry,
+    update_operation::UpdateOperation,
+};
+use crate::model::traversal::state::state_variable::StateVar;
+use itertools::Itertools;
+use serde_json::json;
+use std::collections::HashMap;
+
+pub struct StateModel(HashMap<String, StateModelEntry>);
```
nreinicke, line 10: This can be a future work idea but I tested out storing 3 state model entries as a vector and then just scanning the vector to find the right entry and it's about 4x faster to get the last entry and 8x faster to get the first entry (over 1M iterations). I wonder if we could either default to storing the entries as a vector that we scan or expand the internal store such that if we have a large number of state variables (maybe >10) than we use a hash map but if we have a small number than we use a vector.

rust/routee-compass/src/app/compass/compass_app.rs
```diff
@@ -142,13 +144,18 @@ impl TryFrom<(&Config, &CompassAppBuilder)> for CompassApp {
             .try_deserialize::<serde_json::Value>()?
             .normalize_file_paths(&"", &root_config_path)?;
 
-        let alg_params = config_json.get_config_section(CompassConfigurationField::Algorithm)?;
+        let alg_params =
+            config_json.get_config_section(CompassConfigurationField::Algorithm, &"TOML")?;
         let search_algorithm = SearchAlgorithm::try_from(&alg_params)?;
 
+        let state_params =
```
nreinicke, line 151: I wonder if we could nest the state parameters within a toml section? I think this setup would require us to put the `state` key at the top of the toml file since if it's not then it will get nested inside another section. For example if we have the config:

```toml
[graph]
edge_list_input_file = "../data/tomtom_denver/edges-compass.csv.gz"
vertex_list_input_file = "../data/tomtom_denver/vertices-compass.csv.gz"
verbose = true

state = [
    { distance_unit = "kilometers", initial = 0.0 },
    { time_unit = "minutes", initial = 0.0 },
]
```

state would get nested under the `graph` key. Maybe the config could look something like:

```toml
[state_model]
state = [
    { distance_unit = "kilometers", initial = 0.0 },
    { time_unit = "minutes", initial = 0.0 },
]
```

nreinicke (commented):
This looks great! I like how it centralizes the state information and makes it more intuitive. I left a few comments that we can discuss but overall this looks really good.

rust/routee-compass-core/src/algorithm/search/edge_traversal.rs
```diff
@@ -43,44 +40,53 @@ impl EdgeTraversal {
...
-                            let cost = um
-                                .access_cost(prev_edge, edge, prev_state, &next_state)
-                                .map_err(SearchError::CostError)?;
-                            Ok((next_state, cost))
-                        }
-                        None => Ok((prev_state.to_owned(), Cost::ZERO)),
-                    })
-            })
-            .unwrap_or(Ok((prev_state.to_owned(), Cost::ZERO)))?;
+            let ac = si
```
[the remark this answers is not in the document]
robfitzgerald: we do compute access cost before traversal cost, line 73 then line 89 respectively. 😬

rust/routee-compass-core/src/model/state/state_model.rs
robfitzgerald: totally. that's how [Scala implements `Map`. for Map instances with 4 or less elements](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Map.scala#L259), they rely on a scan, and don't actually instantiate a hashmap until 5+. that could make sense, and perhaps we pull it off by turning StateModel into:

```rust
pub enum StateModel {
  OneFeature {
    name: String,
    entry: StateModelEntry
  },
  TwoFeatures { 
    name1: String,
    name2: String,
    entry1: StateModelEntry,
    entry2: StateModelEntry,
  },
  ...
  NFeatures(HashMap<String, StateModelEntry>),
}
```

rust/routee-compass/src/app/compass/compass_app.rs
robfitzgerald: ya, you're right, and i thought this was awkward too. i like your suggestion and i think it's a simple change.

rust/routee-compass-core/src/algorithm/search/edge_traversal.rs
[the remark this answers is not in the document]
nreinicke: oh lordy I totally misread that, you're right!

rust/routee-compass-core/src/model/state/state_model.rs
nreinicke: yeah that's a good idea, and then getting it would be simple and performant like:

```rust
fn get(&self, name: &str) -> StateModelEntry {
 match self {
   StateModel::OneFeature => self.entry,
   StateModel::TwoFeatures => {
    if self.name1 == name {
     self.entry1
    } else {
     self.entry2
    },
    // and so forth
   }
 }
}
```

robfitzgerald:
@nreinicke GOT IT. (i think)

nreinicke (approved):
Cool, I'm down to bring these changes in. I built some new issues: #137 and #138 based on our discussions in this PR.

PR #150, opened 2024-03-14
fix state model update ordering

### LINKED ISSUE #143: CostModel needs latest StateModel when serializing summary
it appears that the cost model summary only knows about StateFeatures set at config time, because it will produce a cost summary that lacks energy values:

```json
{
  "distance": 1.234,
  "time": 5.678
}
```

this is because the cost model is built at line 168 of search_app.rs, which is before the state model has been finalized. simply moving the CostModelService::build method down and passing it the updated state model will fix this.

robfitzgerald:
this itty bitty PR feeds the updated state model into the builders for CostModel and FrontierModel instead of using the instance of the state model that does not know about TraversalModel features. flipping this ordering fixes the problem where the cost summary doesn't seem to report cost summaries for all features.

Closes #143 .


nreinicke (approved):
Cool, this makes sense. I've tested it and am now seeing all the state variables in the result.

PR #151, opened 2024-03-15
Rjf/unit create by ref

robfitzgerald:
this PR makes it so any units functions take references to their arguments instead of pass-by-value semantics, which should reduce a bunch of unnecessary allocations in our core floating point operations.

still, looking at the bottom of the call stack at the code in routee-compass/src/model/unit/builders.rs, i'm not convinced that we have made these operations as fast as they can be. some additional work could be done to ensure we're not allocating temp objects and perhaps bypassing a conversion to the base units too.


nreinicke (approved):
This makes sense to me, did you notice any performance improvements after implementing this?

robfitzgerald:
Didn't run a test to see, I kinda just whipped this up while on a call with tech support for the VPN problems I've been having, and _assumed_ it would help things. 🤷

PR #153, opened 2024-03-18
Rjf/output name collisions

### LINKED ISSUE #141: fix key collisions during output serialization in the info object
as described in #139, it appears that the new state model serialization is colliding with the existing traversal model serialization during output processing. this can be avoided by targeting a different JSON object for state model serialization.

robfitzgerald:
this PR changes how cost + state outputs are added to the output object. the state model and traversal summary were both using the feature names as keys, causing a key collision. to fix this and improve readability, the values and info sections were unbundled for both cost and state:

```json
{
  "cost": {},
  "cost_model": {},
  "traversal_summary": {},
  "state_model": {}
}
```

additionally, the cost model output was changed from a struct of vecs to nested structs so that all information related to a single feature is organized.

Closes #141.


nreinicke (approved):
Just tested this out. Looks good.

PR #156, opened 2024-03-19
update example notebook; bugfixes

nreinicke:
Supersedes #131 by merging that notebook with the notebook on the main branch.

Also includes a few fixes found along the way:
 - injects distance from the time model until we have a distance model in the scope for the energy traversal model
 - fixes the units for the speed coming out of the time model and into the energy model.

### DIFF
--- a/python/nrel/routee/compass/io/generate_dataset.py
+++ b/python/nrel/routee/compass/io/generate_dataset.py
@@ -73,13 +73,13 @@ def generate_compass_dataset(
     agg = agg if agg is not None else np.mean
 
     # pre-process the graph
-    log.info("processing graph topology and speeds")
+    print("processing graph topology and speeds")
     g1 = ox.utils_graph.get_largest_component(g)
     g1 = ox.add_edge_speeds(g1, hwy_speeds=hwy_speeds, fallback=fallback, agg=agg)
     g1 = ox.add_edge_bearings(g1)
 
     if add_grade:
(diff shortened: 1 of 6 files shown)


robfitzgerald:
cool, thanks for superseding. did you look into whether grade issues were a problem here too, or is that just elsewhere?

nreinicke:
I did double check grade in this case and everything looked okay, (we download the grade as decimal in this case)

python/nrel/routee/compass/io/generate_dataset.py
```diff
@@ -73,26 +73,26 @@ def generate_compass_dataset(
     agg = agg if agg is not None else np.mean
 
     # pre-process the graph
-    log.info("processing graph topology and speeds")
+    print("processing graph topology and speeds")
```
robfitzgerald, line 76: but what about logging? maybe we just need to be sure to set up logging.basicConfig correctly in the python library?

robfitzgerald (approved):
dig it. fun added explorations! i want to defend logging but don't want to get in the way of progress. 🥂

python/nrel/routee/compass/io/generate_dataset.py
nreinicke: yeah there's probably the _right_ way to do it that I didn't have the time to investigate; the motivation was that it was not logging to the jupyter notebook and so it was hard to tell what was happening.

PR #157, opened 2024-03-21
incoming grade unit required

nreinicke:
Update the expected key for the incoming energy model grade units to match the existing format. Also makes the grade unit a required attribute so it doesn't default to something unexpected.


robfitzgerald (approved):
🚗

PR #159, opened 2024-03-22
access models in osm TOMLs

robfitzgerald:
quick tack-on here, i forgot to update the OSM default TOML files with the turn delay access model.


nreinicke (approved):
oh yes, good catch!

PR #160, opened 2024-03-22
remove deprecated key

robfitzgerald:
ugh, one more quick fix, forgot to remove the headings file key from the energy model (and, it blows up when the path normalization function tries to normalize it)


nreinicke (approved):
oooh right, makes sense

PR #163, opened 2024-03-25
state in toml is optional

robfitzgerald:
this quick PR makes it so a `state` section in the TOML is not required. it can still be provided, but is assumed to be an edge case.


nreinicke (approved):
yep, that makes sense

PR #173, opened 2024-04-01
Rjf/specialized indexed map

### LINKED ISSUE #158: StateModel::extend should not warn, but instead, test equality on name collision
extend currently has pretty passive semantics on name collisions (see [here](https://github.com/NREL/routee-compass/blob/main/rust/routee-compass-core/src/model/state/state_model.rs#L127)). but we want all models to report the features they use. in the case that two models expect the same feature, we should require that their StateFeature objects match via an equality test, with one of two outcomes:
  - they match: do nothing
  - no match: return an Err describing the mismatch

### LINKED ISSUE #169: Use Specialized Map from CostModel in Adjacency Matrix
The specialized Map type used in CostModel could also be applied to the HashMap<EdgeId, VertexId> maps in the adjacency matrix. Based on Nick's exploration, this gives a runtime improvement over basic HashMap for small HashMaps, and since that's the case with road networks (out edges are typically on the order of 2-4), we should see a large benefit from this.

Bonus points: create a generic enum SpecializedHashMap<K: Hash, V> that implements these optimizations for any arbitrary K, V.

robfitzgerald:
This PR introduces a generic class modeling a compact hash map. It takes the idea started in the StateModel and builds a generic class around the concept. 

- StateModel is re-implemented with the compact hashmap
- DirectedGraph's adjacency lists now use the compact hashmap

Based on Nick's cursory investigations the latter should help us with runtimes on graph lookups.

Closes #158.
Closes #169.


robfitzgerald:
> request had error: new state features overwriting existing

lol, yes. so, while i was in there i re-wrote the logic to address #158. my mistake was a `==` should have been a `!=`. fixed.

> And running on the compass_tomtom_denver_energy_interpolate.toml config with this query:

that turns out to be correct. the query you are using should replace "state_variable_coefficients" with "weights". also, the config file you mentioned is providing a default set of cost weights, which is what is being read:

```toml
[cost.weights]
distance = 1
time = 1
energy_liquid = 1
energy_electric = 1
```

the state model gets this and gets confused because the model_type for the query is a BEV, which does not have energy_liquid. maybe there's a less _sensitive_ way to implement this that doesn't have gnarly externalities, but this does seem correct for today.

nreinicke:
Yep, good catch on needing to switch to the `weights` keyword. I just tested this and it ran successfully on the denver tomtom test. Strangely I didn't notice any performance boost on the sample denver run.. Here's the result from the main branch:

image

and here's the result from this branch:

image

maybe we'll only see the improvements on large searches or perhaps there is something happening that is eroding the gains like string cloning...

nreinicke (approved):
Thinking it makes sense to merge this in even if we don't notice performance gains on the smaller test cases since it's a nice improvement to the state model. Thanks for making these mods!

PR #174, opened 2024-04-02
flush file sink per optional config parameter

robfitzgerald:
this PR addresses the fact that the line-writing of responses to a file output sink does not have an auto-flush cadence. for a file output policy, the user can now specify an optional `file_flush_rate` which is a positive integer. after writing every `$file_flush_rate` lines, flush will be called on the output buffer. if not specified, flush is called after every row.


nreinicke (approved):
Good find on this one, hopefully it helps alleviate some of the memory pressure

PR #176, opened 2024-04-03
cost attrs in TOML should be optional

robfitzgerald:
ran into this error:

```
Error: CompassConfigurationError(ExpectedFieldForComponent("weights", "cost"))
```

this is because a TOML didn't include a section for cost weights. but, for this server, it was expected that users would always provide these weight values. the CostModelBuilder doesn't allow for optional arguments.

this PR changes that behavior to the opposite extreme, allowing all arguments to be optional. the belief is that the eventual failure is handled downstream in the CostModel::new method, when both TOML and query-time arguments have both been specified.


nreinicke (approved):
Yep that makes sense to me

PR #179, opened 2024-04-04
Rjf/batching cli input

### LINKED ISSUE #117: A streaming "run" application
the current command line, Rust, and Python versions of CompassApp runners expects the full collection of JSON queries as an input, which requires loading them all in memory simultaneously. to improve management of the application memory, we could create an alternative workflow which streams from an input with some given batch size. 

also supports building a web API (#42) which would also deal with streaming inputs over a long-running process.

robfitzgerald:
This PR modifies the command line app for Compass in order to support batch chunking of large query collections, via the following:
  - main.rs contents moved into a routee_compass::app::run module
  - modified file sink policy to have append logic
      - some work toward a future scenario where we may support different write modes
  - CLI accepts chunksize parameter
  - input now accepts regular JSON as well as newline-delimited JSON
      - chunksize only works with newline-delimited inputs
  - command line runner has been written so it can be re-used by downstream libraries

Closes #117.


robfitzgerald:
preview of the revised cli help

```
target/release/routee-compass --help
The RouteE-Compass energy-aware routing engine

Usage: routee-compass [OPTIONS] --config-file <*.toml> --query-file <*.json>

Options:
  -c, --config-file <*.toml>   RouteE Compass service configuration TOML file
  -q, --query-file <*.json>    JSON file containing queries. Should be newline-delimited if chunksize is set
  -c, --chunksize <CHUNKSIZE>  Size of batches to load into memory at a time
  -n, --newline-delimited      Format of JSON queries file, if regular JSON or newline-delimited JSON
  -h, --help                   Print help
  -V, --version                Print version
```

nreinicke (approved):
awesome, this is a nice addition and I appreciate the CLI being cleaner, thanks for adding in!

PR #181, opened 2024-04-04
search termination criteria fix

### LINKED ISSUE #180: edge rtree will explore entire graph in degenerate cases
i'm seeing this with MEP runs. i have some hex that's in northeast alaska. i have the edge_rtree accepting only road classes [4,5,6,7] within 1 kilometer (can you see the problem already? i didn't 😸 ). 

...
...
...
...
...

ok, here's what happens:
  1. no links are found within 1 kilometer that have [4,5,6,7] road class
  2. every link outside 1 kilometer _is outside 1 kilometer_ (yay, a tautology)

result: plugin iterates over every edge in in the graph and returns an error not found.

the solution is a re-write to the `search` method in edge_rtree_input_plugin.rs.

robfitzgerald:
this PR addresses the incorrect logic in the edge rtree search, which previously could visit every edge in the graph for degenerative cases where no valid links exist within the tolerance distance.

Closes #180.


nreinicke (approved):
good find!!

Issue #185, opened 2024-04-10
add optional range restrictions for numeric CustomFeatureFormat instances


robfitzgerald:
if a user can overwrite the initial state of charge, they could do so with the wrong unit (ratio vs percent). to avoid this, we could add optional numeric range fields for these types, and use them during encode/decode operations as constraints which can cause StateError responses. for example:

```rust
pub enum CustomFeatureFormat {
    FloatingPoint { initial: OrderedFloat<f64>, min: Option<OrderedFloat<f64>>, max: Option<OrderedFloat<f64>> },
    ...
}

impl CustomFeatureFormat {
    pub fn encode_f64(&self, value: &f64) -> Result<StateVar, StateError> {
        match self {
            CustomFeatureFormat::FloatingPoint { initial: _, min, max } => {
                let min_val = min.unwrap_or(OrderedFloat(f64::MIN));
                if value < min_val {
                    Err(...)
                } else {
                    Ok(StateVar(*value))
                }
            },
            _ => Err(StateError::EncodeError(
                UnitCodecType::FloatingPoint.to_string(),
                self.name(),
            )),
        }
    }
    ...
}
```

testing this

PR #186, opened 2024-04-10
allow query to overwrite state features

### LINKED ISSUE #145: StateFeature registration via config, service.build, user query
In order to more clearly delineate between TraversalModels in a hierarchical arrangement, we can enforce that any TraversalModel "registers" the feature it updates.

This would assert the following conventions + code changes:

1. state features _should not_ be defined at the root of the TOML in a `[state]` section, but instead, _always_ come from traversal model configuration
2. each traversal model implements `TraversalModel::register_features()` (currently `state_features()`) and returns the state features that it _updates_
    - if a traversal model depends on a feature from an inner model (hierarchical traversal models), it **does not** register the feature, but instead, assumes that the features has been registered by the inner model
    - bonus points: the feature names are parameterized in the config JSON used to build a model, allowing us to define the (somewhat arbitrary) feature names at config time
4. StateModel collects these and builds a state representation as currently implemented
5. Remove StateModelService
    - we no longer persist StateModel feature configurations
    - we no longer define a top-level [state] in the TOML with globally-set state features

robfitzgerald:
This PR gives the user the option to overwrite initial values for state features at query time. the user
  - must be overriding an existing state feature by name
  - must match the feature type of the existing (if it was a StateFeature::Distance, it should still be that)
  - can ovewrite the initial value of the feature (go from 0.0 to 10.0 minutes initial time)
  - can also overwrite the enum variant of the unit (go from Kilometers to Meters)

Added issue #185, that would help further protect against invalid overwrites (a percent should still be a percent).

Closes #145.

### DIFF
--- a/rust/routee-compass-core/src/model/state/state_error.rs
+++ b/rust/routee-compass-core/src/model/state/state_error.rs
@@ -12,6 +12,8 @@ pub enum StateError {
     UnknownStateVariableName(String, String),
     #[error("invalid state variable index {0}, should be in range [0, {1})")]
     InvalidStateVariableIndex(usize, usize),
+    #[error("expected feature to have name '{0}' but found '{1}'")]
+    UnexpectedFeatureName(String, String),
     #[error("expected feature to have {0} unit type but found {1}")]
     UnexpectedFeatureUnit(String, String),
     #[error("{0}")]
--- a/rust/routee-compass-core/src/model/state/state_error.rs
+++ b/rust/routee-compass-core/src/model/state/state_error.rs
@@ -12,9 +12,9 @@ pub enum StateError {
     UnknownStateVariableName(String, String),
     #[error("invalid state variable index {0}, should be in range [0, {1})")]
     InvalidStateVariableIndex(usize, usize),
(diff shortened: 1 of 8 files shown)


nreinicke:
Yeah I like this feature and think it will be important for things like initial state of charge. When attempting to test this, I'm noticing that the units are getting set appropriately when passing in a custom state feature but I'm not seeing the initial value being updated. For example, if I add the following to our base query in the open street maps example notebook:

```json
        "state_features": {
            "time": {"time_unit": "seconds", "initial": 6000}
        },
```

The resulting time comes out as 653 seconds:

image

If I don't inject this the time gets reported as 10.89 minutes and so it's using the right units but not the new initial value.

robfitzgerald:
> it's using the right units but not the new initial value

what's weird is the state model _does reflect the new initial value_ in the screenshot above. wondering where this is getting ignored or overwritten. i think i need to check the initial_state function.

nreinicke (approved):
good find! just pulled the latest changes and it seems to be working as expected, thanks for adding this in!

PR #191, opened 2024-04-19
Rjf/edge rtree optional road class

### LINKED ISSUE #188: edge rtree road class restrictions should be optional
most of the logic in the EdgeRtreeInputPlugin is designed around road class matching being _optional_. however, the builder expects that the road class file is not optional and eagerly builds a lookup table from edge to road class. this should also be optional.

robfitzgerald:
this PR makes it so the road classes argument for an EdgeRtreeInputPlugin can be _optional_. this allows Compass to be configured with edge map matching that does not perform road class filtering.

Closes #188.


nreinicke (approved):
yeah I like this, it's a lot cleaner, thanks for adding in!

PR #192, opened 2024-04-22
K-Shortest Paths Solver

robfitzgerald:
Here is a draft of a ksp algorithm that could possibly work well at national scale. It requires running a bi-directional a* search tree algorithm and then selecting candidate spots from the intersecting vertices from where to join paths from each search. It does this until it can generate $k$ paths meeting some (dis-) similarity threshold.

This _compiles_ but is not correct, but since my parental leave is imminent, I want to surface this in the condition it's in. Planning to wrap it up in the next few days. My current issue is a "search tree is missing linked vertex 413760" error, which may be due to how I'm building the two a* searches (one must be _reverse_-oriented) or how i collect the two routes into a new one.

In order to rule out mapping from edges to vertices, I am simply using the vertex rtree plugin and turning off road class filtering, since those could also cause connectivity issues.

### DIFF
--- a/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
+++ b/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
@@ -9,7 +9,7 @@ use crate::model::road_network::vertex_id::VertexId;
 use crate::model::unit::cost::ReverseCost;
 use crate::model::unit::Cost;
 use crate::util::priority_queue::InternalPriorityQueue;
-use priority_queue::PriorityQueue;
+
 use std::collections::HashMap;
 use std::time::Instant;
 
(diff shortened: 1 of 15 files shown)


robfitzgerald:
@nreinicke i made this a real thing because i'm getting ksp results:

Screenshot 2024-04-23 at 4 33 33 PM

but i forgot to do an important update at the end that propagates the state+cost from the first leg to the second. so, it's not actually ready :blush:

robfitzgerald:
@nreinicke here we go, results from a k=30 search where the threshold for the similarity function was set to 90%:

Screenshot 2024-04-24 at 3 43 31 PM

image

robfitzgerald:
oh, here's time, since that could be more meaningful at a glance 

Screenshot 2024-04-24 at 3 50 30 PM

nreinicke:
😲 Wow! This is amazing! So cool that you were able to put together so fast. I'm excited to review this but in the meantime the results look really promising

nreinicke (approved):
Sorry it took me a while to get to this but it all looks great to me! I've tested this out and everything looks great so far, really nice work!

robfitzgerald:
Sure, thanks for taking the time to review. I'm sick today and missed Friday due to VPN issues, but hope to be back tomorrow, possibly remote (no baby yet)

PR #194, opened 2024-04-26
Add distance to state model in SpeedTraversalModel and other small fixes

nreinicke:
Some small fixes uncovered during a recent debugging session:
 - speed traversal model didn't add distance during the state estimation
 - adds a test to priority queue to make sure it works properly with reverse costs
 - corrects the expected keys for energy time and distance units
 - for the CLI app, sets `--chunksize` to only a long argument since the short argument `-c` conflicts with `--config-file`


robfitzgerald (approved):
🎸

PR #198, opened 2024-04-30
Rjf/a star weight factor

### LINKED ISSUE #195: A* Heuristic Weight Factor
When users run A* searches, it should be possible to apply a global weight factor to all heuristic costs:

```rust
pub enum SearchAlgorithm {
  AStarAlgorithm {
    weight_factor: Option<f64>
  }
}
```

and so SearchAlgorithms should be able to apply heurstic weight factors:

```rust
impl SearchAlgorithm {
  pub fn apply_weight_factor(&self, cost_estimate: Cost) -> Result<Cost, SearchError> {
    ...
  }
}
```

in order to wire this in, SearchAlgorithm would need to be in scope of SearchInstance::estimate_traversal_cost method.

robfitzgerald:
This PR adds an optional argument for the a* algorithm. this allows the user to get dijkstra's algorithm:

```toml
[algorithm]
type = "a*"
weight_factor = 0.0
```

actually, config-wise, there is also a `dijkstra` algorithm now that is semantically equivalent:

```toml
[algorithm]
type = "dijkstra"
```

Closes #195.


nreinicke (approved):
excellent, thanks for adding that in!

Issue #205, opened 2024-05-09
Update cost model weight behavior


nreinicke:
The current behavior for the cost model is to allow users to specify weights at runtime but [fail ](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/rust/routee-compass/src/app/compass/config/cost_model/cost_model_service.rs#L69)if any of the state variables that those weights correspond to are not in the current traversal model. For example, if we load the 2016 Toyota Camry model and then we pass a query that looks like this:

```json
{
  "weights": {"time": 1, "energy_liquid": 1, "distance": 1, "energy_electric": 1}
}
```

We'll get an error back saying there are `unknown state variables in query:`.

This is helpful behavior to make sure we're not mistakenly passing a weight with a typo (which silently reverts to a default) but, when doing a grid search, it is undesirable since we might want to search over multiple vehicle models and our current grid search mechanism doesn't allow us to specify different weights for different vehicles.

So, perhaps we provide a new configuration parameter that defines the behavior for setting weights, something like `expect_exact_state_variables`, where if it's set to true, we get the current behavior but if it's set to false, we allow extra weights to exist in the query but then we just don't use them.

PR #206, opened 2024-05-09
Allow extra weights in Cost Model

nreinicke:
Implements #205 by adding a new configuration parameter on the cost model `expect_exact_weights`. If this is set to false, the application will allow extra weights to be passed through, opening up the possibility of doing a grid search over traversal model that have different weights.


robfitzgerald (approved):
Dig it. Looks good to me. 
Suggestion, and potential nitpick, but, I don't think "expect exact weights" is clear. Maybe "ignore unknown|unregistered feature weights" (with reversed logic) may be more clear? Still a mouthful 😆

nreinicke:
>  Maybe "ignore unknown|unregistered feature weights" (with reversed logic) may be more clear?

I like that idea, I just updated the name to be `ignore_unknown_user_provided_weights` which is hopefully more explicit.

PR #224, opened 2024-05-20
Add optional folium_map parameter to plot_routes_folium.

### LINKED ISSUE #199: Pass optional folium map to plot
Similar to the [`plot_route_folium`](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/python/nrel/routee/compass/plot/plot_folium.py#L16) function where we can pass an optional pre-existing folium map, we should also allow a user to pass a folium map the the [`plot_routes_folium`](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/python/nrel/routee/compass/plot/plot_folium.py#L122) function so it can be customized before plotting the routes on it

vbmendes:
Closes #199.

This PR adds the `folium_map` parameter to the [plot_routes_folium](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/python/nrel/routee/compass/plot/plot_folium.py#L122) function.

For this, I refactored the `plot_route_folium` and broke it into 3 helper functions:
1. `result_dict_to_coords` to convert a result from `CompassApp` into a sequence of coordinates in the format `(lat, lng)`.
2. `_create_empty_folium_map`
3. `plot_coords_folium` to plot the coordinates in a folium Map.

This way, I can reuse these three pieces of code in the `plot_routes_folium` function.


nreinicke (approved):
This looks great, thanks for adding this in!

Issue #216, opened 2024-05-18
Streamline process for loading python CompassApp


nreinicke:
Our current workflow for building an OSM based RouteE Compass app is to create a new osmnx networkx graph and then convert that into our file formats, saving everything into a folder that later gets loaded as a CompassApp object. We could offer a streamlined option that goes straight from a primitive into a fully loaded compass app, something like:

```python
app = CompassApp.from_place("Denver, Colorado")
```

```python
app = CompassApp.from_polygon(some_shapely_polygon)
```

Where, under the hood we call the respective osmnx methods to build the networkx graph, cache all the necessary files and then load the application from the resulting files. Maybe we could use the [`osmnx.cache_folder`](https://osmnx.readthedocs.io/en/stable/user-reference.html#module-osmnx.settings) location to store the files for the application.

nreinicke:
Closed by #225

PR #225, opened 2024-05-20
216 streamline compass app

zenon18:
Completed issue #216

Added two class functions and a helper function to create a CompassApp using a place or a polygon. The helper function will allows users to use a cache directory or use TemporaryDirectory


robfitzgerald:
@nreinicke this makes osmnx a required dependency, should it be optional?

nreinicke (approved):
This looks great, thanks for adding in!

PR #226, opened 2024-05-20
Add Ruff as a linter and formatter, replacing Black

### LINKED ISSUE #215: Look into adding ruff
We currently format the code with `black` but do not enforce any linting. We could consider using ruff for both formatting and linting, adding it as a dev dependency and then enforcing it's checks and formatting with [their github action](https://github.com/chartboost/ruff-action).

BradonZhang:
This change adds Ruff linting and formatting to the GitHub Actions `python-build-test.yaml` workflow, replacing Black as a formatter. This change also applies the necessary linting and formatting to all existing Python files.

Closes #215.

### DIFF
--- a/.github/workflows/python-build-test.yaml
+++ b/.github/workflows/python-build-test.yaml
@@ -40,9 +40,13 @@ jobs:
       # - name: Run mypy
       #   run: mypy . --ignore-missing-imports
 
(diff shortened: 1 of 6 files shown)


nreinicke (approved):
Thank you for adding this in!

PR #227, opened 2024-05-20
Complete type annotations

### LINKED ISSUE #214: Finish adding mypy type checking
We've started adding type hints to the [python wrapper](https://github.com/NREL/routee-compass/tree/main/python/nrel/routee/compass) of the app but haven't gotten full coverage. Add in any missing types and then we can activate the [mypy check ](https://github.com/NREL/routee-compass/blob/4a764088d0372a08596c274ed90f4bf4b884dad7/.github/workflows/python-build-test.yaml#L40) in our GitHub action.

SnoopJ:
Closes #214

This changeset adds a bunch of missing annotations and other fixes/workarounds to get `routee-compass` to a complete set of annotations (i.e. `mypy --strict` can be run over the whole wrapper with no errors). This also updates and re-enables the type-checking step of the test workflow (and adds the necessary type stubs to the `[dev]` extra)

### DIFF
--- a/.github/workflows/python-build-test.yaml
+++ b/.github/workflows/python-build-test.yaml
@@ -37,8 +37,8 @@ jobs:
         run: |
           pip install ".[dev]"
 
-      # - name: Run mypy
-      #   run: mypy . --ignore-missing-imports
+      - name: Run mypy
+        run: mypy --strict ./python/nrel
 
       - name: Validate formatting
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,17 @@ classifiers = [
 keywords = ["eco routing"]
 dependencies = ["toml"]
 [project.optional-dependencies]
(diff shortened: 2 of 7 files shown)


nreinicke:
Thanks for adding all of these in! We made a couple of changes when sprinting today that will cause a couple merge conflicts here (sorry!)

We also decided to make the geospatial-like dependencies optional and then just rely on the `TYPE_CHECKING` parameter and en-quoting the optional types.

SnoopJ:
No trouble at all on the merge conflicts, I'll sort them out. Guarding/quoting the geospatial stuff sounds pretty straightforward to take care of.

It turned out that I didn't have much time today, but I did get a few moments to look at the problems with the annotation of the `agg` parameter. The core problem is I believe the conflict between `Union[VeryLongCallableT, overloaded function]` and `Optional[VeryLongCallableT]`, but I'm not entirely sure what the overloaded bit is about.

Issue #209, opened 2024-05-14
Add batch running capability to python API


nreinicke:
The current workflow when using the python API is to load a `CompassApp` object from a config and then call the `run()` method, passing a set of queries as python dict objects. We could also add another way to interact with the application which would be to point to a config file and a query json file and then have the application read the queries in batches and also write to an output file in batches. This could take the form of a single function that looks like this:

```python
def run_in_batches(app_config_file: Union[str, Path], query_file: Union[str, Path], batch_size: int):
    ...
```

robfitzgerald:
and assuming we could wire this in via the CompassApp.run method:

```python
class CompassApp:

  def run(..., chunksize: Optional[int]):
      if chunksize is not None:
          return run_in_batches(...)
```

thinking: i feel like stacking alternatives into the core run method would probably be more easily discoverable for users. same goes for the "to geopandas" behaviors, and, maybe all of that wiring-in is just one more downstream issue to collect these all as behaviors of `CompassApp.run`, which gives us one of those big APIs like `Pandas.read_csv` or similar that serves all users.

nreinicke:
> maybe all of that wiring-in is just one more downstream issue to collect these all as behaviors of CompassApp.run, which gives us one of those big APIs like Pandas.read_csv or similar that serves all users.

Yeah I like that idea. I changed to scope of this to just complete the function and then we can build out a new issue for wiring it in (along with any other run methods we discover)

zenon18:
zenon18 (Mark W.) will like to work on this issue

PR #228, opened 2024-05-21
209 run in batches

zenon18:
Pull request for #209

Created a generator function that takes a file pointer and a batch size and reads up to that many queries from the file and calls run.

### DIFF
--- a/python/nrel/routee/compass/compass_app.py
+++ b/python/nrel/routee/compass/compass_app.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional, Union, Generator
 from nrel.routee.compass.routee_compass_py import (
     CompassAppWrapper,
 )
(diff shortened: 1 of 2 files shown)


python/nrel/routee/compass/compass_app.py
```diff
@@ -135,6 +135,55 @@ def run(
             return results[0]
         return results
 
+    def run_in_batches(
+        self, fp: TextIO, batch_size: int
```
jhoshiko, line 139: Instead of passing a file pointer, let's instead pass a Path or string and then create the file pointer in this method.

zenon18: gotcha will update

jhoshiko:
After discussing this feature with the group, it sounds like we should look into adding a parser for query files to prevent crashes from faulty files (See issue # 236)

PR #230, opened 2024-05-21
Ndr/result to geopandas

### LINKED ISSUE #210: route_to_geopandas function
To support #200, we should create a new function `route_to_geopandas` that converts the `"route"` key of the result (python dictionary) into a geopandas dataframe with the following information:

<h4 dir="auto">Summary row data</h4>
<p dir="auto">Any row in a route GeoDataFrame should have all traversal and cost columns that may be relevant to post-processing:</p>

column | mapping
-- | --
_index | (user-defined index_col path, or, by default, the top-level index value in the Compass output)
distance | traversal_summary.distance
time | traversal_summary.time
... |  
distance_cost | cost.distance
time_cost | cost.time
total_cost | cost.total_cost

Summary row data

Any row in a route GeoDataFrame should have all traversal and cost columns that may be relevant to post-processing:
column 	mapping
_index 	(user-defined index_col path, or, by default, the top-level index value in the Compass output)
distance 	traversal_summary.distance
time 	traversal_summary.time
... 	
distance_cost 	cost.distance
time_cost 	cost.time
total_cost 	cost.total_cost

All of these should be set only if they exist (optional semantics).
Geo row identifiers

The route geometries are parsed as geometry objects. A _path column is added to the dataframe in the case that there is more than one path. The value is the route.path index.
run() result 	description 	route column?
one route {} 	route.path has one entry 	no
n routes [] 	route.path has i entries 	append _path column with value i

As a result, a row is uniquely indexed by the _index column when there is one route per row, and a combination of _index and _path when route.path has more than one entry. Put another way: we end up with k * q rows for q queries and k paths.
Geometry data

We need to cover the possible geometry output types. We can know in advance what the geometry type is by inspecting the configuration of the CompassApp (via https://github.com/NREL/routee-compass/issues/201). We can then switch our geometry parsing method accordingly:
Compass traversal.route config value 	parse method for string s
json 	(no geometry)
wkt 	shapely.wkt.loads(s)
geojson 	shapely.ops.shape(s)

### LINKED ISSUE #211: tree_to_geopandas function
To support #200 we can create a new function `tree_to_geopandas` function that converts the `"tree"` key from the default result (python dictionary) into a geopandas dataframe.

Each Compass output row may have an entire tree, but, we do not want a GDF of _trees_ (plural), we want instead to plot a single tree. Following the logic above, we can add one more index column so that each row is a link in a tree in a result:

```
for each Compass run result row _i_:
  for each tree _j_:
    for each tree branch _k_:
      create row for `_index` _i_, `_tree` _k_,  `_edge_id` edgeid(_k_)
```

nreinicke:
Adds support for converting the results to geopandas. Now you can call the function `results_to_geopandas` and get back either a route geodataframe (if there is just a route), a tree geodatarame (if there is just a tree), or both a route and a tree geodataframe. For example:

```python
route_gdf, tree_gdf = results_to_geopandas(result)
```

Closes #211 
Closes #210

### DIFF
--- a/python/nrel/routee/compass/compass_app.py
+++ b/python/nrel/routee/compass/compass_app.py
@@ -13,8 +13,9 @@ import toml
 
 
 Query = Dict[str, Any]
-Result = List[Dict[str, Any]]
-
+Result = Dict[str, Any]
+Results = List[Result]
+Route = Dict[str, Any]
 
 log = logging.getLogger(__name__)
 
@@ -89,7 +90,7 @@ class CompassApp:
 
     def run(
         self, query: Union[Query, List[Query]], config: Optional[Dict] = None
-    ) -> Result:
+    ) -> Union[Result, Results]:
         """
         Run a query (or multiple queries) against the CompassApp
 
--- a/python/nrel/routee/compass/io/convert_results.py
+++ b/python/nrel/routee/compass/io/convert_results.py
@@ -1,0 +1,96 @@
+from typing import Optional, Tuple, Union
+
+from nrel.routee.compass.utils.geometry import geometry_from_route
+import pandas as pd
+import geopandas as gpd
+from nrel.routee.compass.compass_app import Result, Results
+
+
+def tree_result_to_geopandas(
+    result: Result,
+) -> Optional[gpd.GeoDataFrame]:
+    """ """
+    if "error" in result:
+        raise ValueError(f"Error in result: {result['error']}")
+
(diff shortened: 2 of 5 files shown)


jhoshiko:
@nreinicke Nice! I've tested this on my end and can generate the tree gdf and the results gdf from both a list of results and a single result.

jhoshiko (approved):

Issue #201, opened 2024-05-09
Store config object in CompassApp instance


robfitzgerald:
we create a CompassApp using a configuration object or file. in both cases, it would be a simple step to store the configuration as an object on our CompassApp so that it can be inspected at a later time.

zenon18:
zenon18 (Mark W.) will like to work on this issue

nreinicke:
Closed by #231

PR #231, opened 2024-05-21
config is now stored in compass app

zenon18:
#201 completed

The config is now stored in self._config


jhoshiko:
Verified that CompassApp stores the config file contents.

jhoshiko (approved):

Issue #213, opened 2024-05-16
Enable adding grade via google API


nreinicke:
Our current workflow for adding grade to the graph is to download USGS raster elevation tiles and then add them to the osmnx graph using the [`add_node_elevations_raster`](https://github.com/NREL/routee-compass/blob/4a764088d0372a08596c274ed90f4bf4b884dad7/python/nrel/routee/compass/io/utils.py#L169) function. osmnx also support the [google maps elevation api ](https://osmnx.readthedocs.io/en/stable/user-reference.html#osmnx.elevation.add_node_elevations_google)and so we could add an additional pathway for our users who have an API key to google maps to use that functionality as well.

zenon18:
zenon18 (Mark W.) will like to do this issue

PR #232, opened 2024-05-21
updated add_grade_to_graph to allow users to pass in a google api key

zenon18:
#213 completed

Updated add_grade_to_graph to have a api_key parameter


nreinicke (approved):
This look good to me

PR #233, opened 2024-05-21
Add OSM/Geospatial dependencies as default

nreinicke:
Previously we only had the OSM/Geospatial dependencies required in the conda install but now that we're adding default functionality that use these dependencies (osmnx, geopandas, etc.), I've included them as required dependencies in the `pyproject.toml` file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,17 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 keywords = ["eco routing"]
-dependencies = ["toml"]
+dependencies = [
+    "toml",
+    "numpy",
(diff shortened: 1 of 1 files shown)


robfitzgerald:
Oh, seeing this one now. I'm not sure I agree with the idea of making these required. Perhaps they could be a feature install, and that feature is the default, but can be deactivated? Osmnx is a heavy weight dependency and it's not actually a requirement for all downstream compass applications (same goes for all of these geospatial libs)

jhoshiko (approved):

jhoshiko:
Discussed this with Nick in person and am approving this change.

PR #234, opened 2024-05-21
usgs using 0 leading numbers in their webpages

zenon18:
Found a bug where some grade maps could not be downloaded, for ex Pittsburgh, PA as it had a two digit longitude and the website requires leading 0s in their longitudes < 100


jhoshiko:
Nice catch! Glad we were able find this bug and patch it.

jhoshiko (approved):

Issue #135, opened 2024-03-07
n-dimensional interpolation models


robfitzgerald:
We currently support [a bilinear interpolation model](https://github.com/NREL/routee-compass/blob/main/rust/routee-compass-powertrain/src/routee/prediction/interpolation/bilinear_interp.rs) for providing very fast energy rate predictions by doing a linear interpolation over a structured grid. This could be extended to support N dimensions such that it could support more than two incoming features (currently just speed and grade). 

Some things to keep in mind is that we want the `interpolate` method to be as fast as possible since it will be called for every link traversal, minimizing any memory allocations and ideally just performing simple operations. 

This work will eventually support a broader refactor of the `PredictionModel` trait that supports an arbitrary number of incoming features, building toward this API:

```rust
pub trait PredictionModel: Send + Sync {
    fn predict(
        &self,
        state: &[StateVar],
        state_model: &StateModel,
    ) -> Result<EnergyRate, TraversalModelError>;
}
```

nreinicke:
@kylecarow - just updated this if you end up having the chance to work on it.

kylecarow:
Working in [`kjc/interpolation`](https://github.com/NREL/routee-compass/tree/kjc/interpolation)

kylecarow:
Compass now uses the new interpolation stuff, but the generic feature support needs some thinking and implementing.

PR #235, opened 2024-05-21
N-dimensional interpolation

kylecarow:
I've plumbed up the 0/1/2/3/N-dimensional interpolation code that I developed for FASTSim to the `routee-compass-powertrain` crate's `ModelType::Interpolate`. This is capable of taking any dimension data, with `Interpolator` enum variants that allow hardcoded interpolation for performance benefits (there are benchmarks in FASTSim 3 for interpolation). It supports multiple interpolation 'strategies' for 1D interpolation, but only linear for other dimensionalities.

There are more improvements to be made, as documented in #135, but this passes all tests in Rust and the single test in Python. Notably, the 2D speed/grade stuff in `InterpolationSpeedGradeModel::new()` is still hardcoded, so more work and thought is required to allow more dimensions and proper mapping of dimensions (and any unit considerations).

See the tests in `interp.rs` for examples on usage. Be sure to call `.validate()` after instantiating to run checks on the supplied data.


rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -33,17 +33,26 @@ impl PredictionModel for InterpolationSpeedGradeModel {
...
 
         // snap incoming speed and grade to the grid
-        let min_speed = self.interpolator.x[0].0;
-        let max_speed = self.interpolator.x[self.interpolator.x.len() - 1].0;
-        let min_grade = self.interpolator.y[0].0;
-        let max_grade = self.interpolator.y[self.interpolator.y.len() - 1].0;
+        let (min_speed, max_speed, min_grade, max_grade) = match &self.interpolator {
+            interp::Interpolator::Interp2D(interp) => (
+                interp.x[0],
+                *interp.x.last().unwrap(),
```
nreinicke, line 39: Can we get rid of these unwraps and map the error into a `TraversalModelError`?

nreinicke: You can add a new error here that is transparent and takes in whatever errors you're dealing with and it will automagically convert it into a `TraversalModelError` if you use the `?` operator.

nreinicke (changes requested):
This is looking great, nice work on making the interpolation so flexible. I left a couple small comments about converting a few of the unwraps into errors.

robfitzgerald (changes requested):
Thanks @kylecarow, excited to see this open new doors for our powertrain models. The code is great, nicely structured. I have a few comments but have only done a quick proofread:
    - following on nicks suggestion, remove _any_ calls to .unwrap() that aren't in a test suite
    - Code was copied from a fastsim pull request, we should attribute it via a code comment at the top of the file

kylecarow:
Done with converting unwraps to results! Also switched to binary search (short circuiting if it's the point is the last element of the grid), for convenience and because it should scale better.

Finished just before my flight boarded, lol.

nreinicke (approved):
Thanks for making those fixes! I've tested this out and it seems to be working as intended.

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -33,17 +33,34 @@ impl PredictionModel for InterpolationSpeedGradeModel {
...
+                interp.x[0],
+                *interp.x.last().ok_or_else(|| {
+                    TraversalModelError::PredictionModel(
+                        "Could not get last x-value; are x-values empty?".to_string(),
+                    )
+                })?,
+                interp.y[0],
+                *interp.y.last().ok_or_else(|| {
+                    TraversalModelError::PredictionModel(
+                        "Could not get last x-value; are x-values empty?".to_string(),
```
kylecarow, line 47:
```suggestion
                        "Could not get last y-value; are y-values empty?".to_string(),
```

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -33,17 +33,34 @@ impl PredictionModel for InterpolationSpeedGradeModel {
...
-        let max_grade = self.interpolator.y[self.interpolator.y.len() - 1].0;
+        let (min_speed, max_speed, min_grade, max_grade) = match &self.interpolator {
+            interp::Interpolator::Interp2D(interp) => (
+                interp.x[0],
+                *interp.x.last().ok_or_else(|| {
+                    TraversalModelError::PredictionModel(
+                        "Could not get last x-value; are x-values empty?".to_string(),
+                    )
+                })?,
+                interp.y[0],
```
robfitzgerald, line 44: Hey guys one least question, sorry for being a bottleneck here. Using the index operator here will cause a panic if the response is not as expected. Shouldn't we be using .get instead?

robfitzgerald (changes requested):
One more thing I noticed that I think we want to change before we roll this in (correct me if I'm wrong, @nreinicke )

rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
nreinicke: Yeah that's a good point. Ideally we would never get to this point if the interpolation grid was empty but it's probably good practice to use get instead (unless that has a performance cost associated with it). Perhaps it would be good enough to just add an additional check into the `validate` function that checks to make sure the grid data always has at least two points and give an instructive error at that point.

kylecarow:
@robfitzgerald @nreinicke I've added additional checks to validation to prevent panics on bad inputs, as well as corresponding tests. I think it would be kind of impossible/impractical to code out *every* use of [] for indexing, things are indexed quite a lot - I have replaced it where possible, and left comments in-code where direct indexing is okay because the data is checked by validation methods. Most of the remaining indexing should be outputs from `find_nearest_index`, so if there are panics with indexing in the future, it means there is likely a bug in that function, which should already be pretty robust.

nreinicke:
Nice, thanks for adding the additional checks in. I'm on board with your assessment about not modifying every index as long as we have the appropriate validation check at load time. I'll leave the floor open to @robfitzgerald for any other comments but overall this looks good to me.

robfitzgerald:
@kylecarow i'm sorry i gotta be a stickler on this since this is in the core lib, so i would prefer we don't pass these potential panics onto users if they do something unexpected. but i think this could be a quick fix, it's a good opportunity for writing helper functions that can encapsulate the error handling. that should split the difference for us wrt conciseness and performance. i may have this wrong but here's a guess on what those signatures would look like:

```rust
fn interp1(f: Vec<f64>, x_diff: f64, l: usize, u: usize) -> Result<f64, String> { todo!() }
fn interp2(f: Vec<Vec<f64>, x_diff: f64, l: (usize, usize), u: (usize, usize)) -> Result<f64, String> { todo!() }
fn interp3(f: Vec<Vec<Vec<f64>>>, x_diff: f64, l: (usize, usize, usize), u: (usize, usize, usize)) -> Result<f64, String> { todo!() }
fn interpN(f: ArrayD<f64>, x_diff: f64, l: Vec<usize>, u: Vec<usize>) -> Result<f64, String> { todo!() }
```

```rust
let c00 = interp3(self.f_xyz, x_diff, (x_l, y_l, z_l), (x_u, y_l, z_l))?; 
```

or if you don't like that, i could also imagine just array lookup helpers like this:

```rust
fn arr3(f: Vec<Vec<Vec<f64>>>, x: usize, y: usize, z: usize) -> Result<f64, String> { todo!() }
```

that could be used like this:

```rust
let c00 = arr3(self.f_xyz, x_l, y_l, z_l)? * (1.0 - x_diff) + arr3(self.f_xyz(x_u, y_l, z_l)? * x_diff; 
```

or anything else you think could make this not be super annoying to implement!

robfitzgerald (approved):
ack, i wish i wasn't on leave, i just can't devote my attention to this the way i would like to, which makes me an evil online PR villain. realizing that i was missing the point here about the fact that validate is called every time. while that does sound expensive i think we are in good shape for now. sorry for being a nuisance!

nreinicke:
No, this is good discussion, you're definitely not being a nuisance! I think the concept is that the `validate` method gets called just once [when the interpolator is getting built](https://github.com/NREL/routee-compass/blob/e48f30e237d315e78dfb435c04f9e701fbdec446/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs#L139) which gives us our guarantees about the index bounds.

robfitzgerald:
> I think the concept is that the validate method gets called just once [when the interpolator is getting built](https://github.com/NREL/routee-compass/blob/e48f30e237d315e78dfb435c04f9e701fbdec446/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs#L139) which gives us our guarantees about the index bounds.

Well, in that case, I would love it if we had a `Interpolator::new` method that calls validate internally. This gets out of the OO-ish framing of "there's a method you need to know gets called first to use this." But also recognizing that the new method would need to take an `ArrayD` and then do the work to specialize that down to 1/2/3D arrays when appropriate in the new method. Could we make an issue to come back to that?

kylecarow:
> validate is called every time. while that does sound expensive

@robfitzgerald 
No worries, here's a description of the data validation that Nick and I had discussed in-person at PyCon:

There are two validation methods:
- The first method validates the data that does not change on calls of the interpolator, i.e. the grid coordinates and corresponding values.
  - Example: checking the coordinates are monotonically increasing, that data isn't empty, etc.
  - This only runs when the interpolator is created, e.g. with `InterpolationSpeedGradeModel::new()`, I imagine only once
    - https://github.com/NREL/routee-compass/blob/e48f30e237d315e78dfb435c04f9e701fbdec446/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs#L134-L144
- The second method validates data supplied in the actual `.interpolate` call, i.e. the interpolant point.
  - Example: checking the point is within the grid, is the applicable length for the dimensionality, etc.
  - This runs whenever the interpolator tries to interpolate at a point, so whenever `InterpolationSpeedGradeModel::predict` from the `PredictionModel` impl is called
    - https://github.com/NREL/routee-compass/blob/e48f30e237d315e78dfb435c04f9e701fbdec446/rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs#L69-L74

Does that help clear up the data validation process? If this still seems expensive I'm open to suggestions for improvements :)

kylecarow:
@robfitzgerald I had wanted an `Interpolator::new` method that called validate, but struggled with it due to the way the interpolator variants need to be instantiated, it would be hard to create an function signature that is clear, concise, and allows passage of the data needed for any variant.

A thought that just occurred to me is to have a `new` method for the variant structs, so that creating a 2D interpolator would change like this:
```diff
- let interp = Interpolator::Interp2D(Interp2D {
-     x: vec![0.05, 0.10, 0.15],
-     y: vec![0.10, 0.20, 0.30],
-     f_xy: vec![vec![0., 1., 2.], vec![3., 4., 5.], vec![6., 7., 8.]],
- });
- interp.validate().unwrap();
+ let interp = Interpolator::Interp2D(Interp2D::new(
+      x: vec![0.05, 0.10, 0.15],
+      y: vec![0.10, 0.20, 0.30],
+      f_xy: vec![vec![0., 1., 2.], vec![3., 4., 5.], vec![6., 7., 8.]],
+  ));
```

We could then restrict users from creating the variant structs directly (i.e. via `Interp2D {...}`) by including a phantom non-`pub` field as described in number 2 here: https://stackoverflow.com/a/77683321/11278044

robfitzgerald:
nice work, @kylecarow. while using `PhantomData` feels a little spooky :ghost: i think it seems it's the smoothest way to hack these default enum constructors to be private and enforce our validation at construction time. nice work! and we can always add the more general `Interpolator::new` method which takes an arbitrary `ArrayD` and specializes down to 1/2/3D at a later time, and it would just use the constructors you wrote. thanks!

kylecarow:
Just pushed a minor commit to remove a superfluous case in output handling from `find_nearest_index` and a very minor formatting tweak. It should be good to merge whenever, so long as you both still approve.

And the notes above about the need to rework some code higher up in the call chain to actually be able to use features other than speed and grade still applies.

PR #237, opened 2024-05-21
fixes types from optional dependency

nreinicke:
Updates the typing slightly to en-quote any type that is from an optional dependency so that it doesn't fail at runtime but still works with mypy.


jhoshiko (approved):
Looks good!

PR #239, opened 2024-05-30
Fix how we download grade; bump version

nreinicke:
Applies some small fixes to how we download grade from USGS. The USGS tile system includes all the points below the tile index. For example, the point (39.5, -104.5) corresponds to the tile n40w105. Previously we were assuming that point would be included in the tile n39w104.


robfitzgerald (approved):
Ah, dig it, exclusive range semantics. Sounds good!

PR #243, opened 2024-07-09
Add section to docs about other vehicle powertrains

nreinicke:
Add an additional section to the open street maps notebook that talks about how to use additional powertrains.


robfitzgerald (changes requested):
nice addition. we're probably overdue for splitting this notebook out into separately-titled (and scoped) example notebooks. 🤷‍♂️ i've got two formatting-ish notes here but that's it.

1. noticed that, in the new section, we're just dumping dicts into the console, but we made that nice `pretty_print` function earlier in the notebook, so we should use that to make it easier to read. 
2. the section title "Other Powertrains" led me to believe this was going to be about RouteE Powertrain integration. maybe "Other Powertrain Types"? or changing the first sentence. right now it says "In this section we'll look at some nuance when running different powertrain types." maybe it should first say something like "RouteE Compass can execute search plans for internal combustion (ICE), plug-in hybrid (PHEV) and battery electric (BEV) powertrain types."

nreinicke:
Yes, great points, I've updated the title to read "Running Other Powertrain Types`, have updated the language a bit and have used the pretty_print methods.

> we're probably overdue for splitting this notebook out into separately-titled (and scoped) example notebooks. 

Yes, we definitely are. We could probably build off https://github.com/NREL/routee-compass/pull/238

robfitzgerald (approved):
🦞

PR #245, opened 2024-07-24
release 0.8.0 (rust)

robfitzgerald:
branched this from the py-0.8.0 tag and updated version references to 0.8.0 in the rust crates in order to release on crates.io as v0.8.0.


nreinicke (approved):
Yep this makes sense, thanks for thinking to do this.

Issue #202, opened 2024-05-09
Compute edge headings for start and end


nreinicke:
In our [current `generate_dataset` function](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/python/nrel/routee/compass/io/generate_dataset.py#L157), we use the edge bearings [generated from osnmx](https://github.com/gboeing/osmnx/blob/d9c05410eed705c482d136ba4750ed83c5c2f778/osmnx/bearing.py#L86) which computes the bearing from the origin node to the destination node. This is a good approximation but RouteE Compass allows us to specify a bearing for both [the beginning and end of a road link.](https://github.com/NREL/routee-compass/blob/fb8376760a7cd4447db71fcc40b193cb0efbcacb/rust/routee-compass-core/src/model/access/default/turn_delays/edge_heading.rs#L7) This is useful for computing the difference in angle between two links that might have significant curvature.

Imagine a (somewhat contrived) example where we have two links with curvature connected to each other like this:

![image](https://github.com/NREL/routee-compass/assets/1743744/ed0c2609-06d6-46f9-821e-c7ab51ee39b6)

If we just use the bearing between the start and end nodes, it will appear as if the vehicle is not turning at all between the two links but in reality the vehicle is making an ~90 degree turn.

In order to support this we should use the geometry from the link and then take the first two points in the linestring to compute the `arrival_heading` and the last two points in the linestring to compute the `departure_heading`.

nreinicke:
Adding a bit more context to this:

We'll need to first extract the edge geometries from the incoming networkx graph. We can get the edge geometries in a geopandas dataframe like this:

```python
edges_gdf = ox.graph_to_gdfs(g, edges=True, nodes=False)
edges_gdf["geometry"]
```

Then, we need to compute the arrival and departure heading for each edge in that dataframe. Here are a couple of functions that we could use:

```python
import math 

def compass_heading(point1, point2):
    lon1, lat1 = point1
    lon2, lat2 = point2

    lat1, lon1, lat2, lon2 = map(math.radians, [lat1, lon1, lat2, lon2])

    dlon = lon2 - lon1

    x = math.sin(dlon) * math.cos(lat2)
    y = math.cos(lat1) * math.sin(lat2) - (
        math.sin(lat1) * math.cos(lat2) * math.cos(dlon)
    )

    initial_bearing = math.atan2(x, y)

    initial_bearing = math.degrees(initial_bearing)
[...]

PR #247, opened 2024-08-13
Add departure headings calculation

nataliejschultz:
In reference to #202:
Previously, generate_dataset would calculate the arrival heading using the preset osmnx `.bearing` method, while departure headings were set to `None`. 

This PR adds two functions written by @nreinicke:
1. `calculate_bearings`, which passes in linestrings and determines what to do with the coordinates based on the number of lat lon pairs
2. `compass_heading`, which calculates a compass bearing based on two lat lon pairs

These functions replace the osmnx method used previously, rather than using osmnx for arrival and the functions for departure. Both headings are added to the same data frame and output to edges-headings-enumerated.csv

### DIFF
--- a/python/nrel/routee/compass/io/generate_dataset.py
+++ b/python/nrel/routee/compass/io/generate_dataset.py
@@ -6,8 +6,8 @@ import importlib.resources
 import logging
 import shutil
 import pandas as pd
-import utils
 
+from nrel.routee.compass.io import utils
 from nrel.routee.compass.io.utils import add_grade_to_graph
 
 log = logging.getLogger(__name__)
--- a/python/nrel/routee/compass/io/generate_dataset.py
+++ b/python/nrel/routee/compass/io/generate_dataset.py
@@ -5,6 +5,8 @@ from pkg_resources import resource_filename
 import importlib.resources
 import logging
 import shutil
+import math
+import pandas as pd
 
 from nrel.routee.compass.io.utils import add_grade_to_graph
 
@@ -154,13 +156,8 @@ def generate_compass_dataset(
         header=False,
     )
 
-    headings = e.bearing.fillna(0).apply(lambda x: int(round(x)))
-    headings_df = headings.to_frame(name="arrival_heading")
-
-    # We could get more sophisticated and compute the end heading
-    # for links that might have some significant curvature, but
-    # for now we'll just use the start heading.
-    headings_df["departure_heading"] = None
+    headings = [calculate_bearings(i) for i in e.geometry.values]
+    headings_df = pd.DataFrame(headings, columns = ["arrival_heading", "departure_heading"])
     headings_df.to_csv(
         output_directory / "edges-headings-enumerated.csv.gz",
         index=False,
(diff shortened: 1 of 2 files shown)


python/nrel/routee/compass/io/generate_dataset.py
```diff
@@ -209,3 +206,38 @@ def replace_id(vertex_uuid):
             with importlib.resources.as_file(model_file) as model_path:
                 model_dst = model_output_directory / model_path.name
                 shutil.copy(model_path, model_dst)
+
+#Function written by Nick Reinicke
+def compass_heading(point1, point2):
```
nreinicke, line 211: Could we move these functions into the [utils](https://github.com/NREL/routee-compass/blob/main/python/nrel/routee/compass/io/utils.py) module?

python/nrel/routee/compass/io/generate_dataset.py
```diff
@@ -209,3 +206,38 @@ def replace_id(vertex_uuid):
...
+    )
+
+    initial_bearing = math.atan2(x, y)
+
+    initial_bearing = math.degrees(initial_bearing)
+    compass_bearing = (initial_bearing + 360) % 360
+
+    return compass_bearing
+
+#Function written by Nick Reinicke
```
nreinicke, line 231: I appreciate the attribution but I'm okay if we omit them 🙂

nreinicke (changes requested):
This is looking great! I left a couple of small comments but I like how simple you've kept this

nataliejschultz:
I added the two functions to `utils.py` and changed the relevant command:

`headings = [utils.calculate_bearings(i) for i in e.geometry.values]`

 For testing these changes locally, I added a name == main if statement to the bottom of the file with parameters to run:

```
if __name__ == '__main__':
    g = ox.graph_from_place("Denver, Colorado, USA", network_type="drive")
    testing_changes = generate_compass_dataset(g, 'testing-terminal')
```
Ran with `python ./routee-compass/python/nrel/routee/compass/io/generate_dataset.py` in terminal. After cd-ing into the testing-terminal output directory and using `gunzip`, here's what the top of the file looks like:

```
(base) nschultz-35728s:testing-terminal nschultz$ gunzip edges-headings-enumerated.csv.gz 
(base) nschultz-35728s:testing-terminal nschultz$ head edges-headings-enumerated.csv
arrival_heading,departure_heading
111,112
139,131
127,122
124,95
91,90
270,270
89,89
0,0
269,269
```
All is well!

python/nrel/routee/compass/io/generate_dataset.py
nataliejschultz: > Could we move these functions into the [utils](https://github.com/NREL/routee-compass/blob/main/python/nrel/routee/compass/io/utils.py) module?

Moved with commit b9f77261ebed959dd01386be240f4b2ee06d9537

nataliejschultz:
@nreinicke [tests failed ](https://github.com/NREL/routee-compass/actions/runs/10377622217) because the runner couldn't find utils, so I changed the `import utils` that worked on my machine to the more universal:

`from nrel.routee.compass.io import utils` 

in commit b6cbd623f0a46db11478fa61bc9262735dcd1658. Hopefully this solves it!

nreinicke (approved):
Excellent, looks good!

PR #249, opened 2024-08-20
Rjf/248 improve ksp

robfitzgerald:
This PR adds some new features to search:
  - [Yen's k-Shortest Paths](https://en.wikipedia.org/wiki/Yen%27s_algorithm) implementation
  - ksp termination criteria to give finer grain control over how many iterations are run
  - revised route similarity function for clearer semantics + call site
  - all search algorithms can receive JSON parameters from the search query 
      - _k_, or A Star _weight_factor_ are the first implementations

submitting this draft for any early review but I still need to test it later this week.


robfitzgerald:
@nreinicke this is ready for review. after running, i was able to explore k=[10,20,30,40,50] and plot which set each result comes from here, colored by it's k value:

pareto-all-kcolor-trip=9541722-model=BEV

i don't have any other tests, so if you feel like this could use a test, let me know.

nreinicke (approved):
This looks good to me, it's cool to see the k paths building out the "cloud" around the Pareto front.

PR #250, opened 2024-08-27
Rjf/wkb

robfitzgerald:
this PR adds WKB read/write operations:
  - write routes + trees in WKB format
  - read from WKB for flat files
    - business logic added, but, not wired in as a configurable alternative

only tested on my laptop, but, it worked:

Screenshot 2024-08-27 at 3 20 13 PM

Screenshot 2024-08-27 at 3 19 50 PM


nreinicke (approved):
Nice! this is a good addition, thanks for adding it in!

PR #252, opened 2024-09-03
dependencies bump

robfitzgerald:
all rust dependencies bumped to latest versions. ran test notebook, some numerical result values are slightly different but past the 6th decimal place, perhaps changes in ndarray or ordered_float.


nreinicke (approved):
dig it

PR #254, opened 2024-09-05
Rjf/soc bug

### LINKED ISSUE #241: Query starting soc not propagating to state model
If I pass a query that looks like this:

```json
[
    {
        "origin_x": -104.969307,
        "origin_y": 39.779021,
        "destination_x": -104.975360,
        "destination_y": 39.693005,
        "model_name": "2017_CHEVROLET_Bolt",
        "starting_soc_percent": 80,
    },
]
```

The resulting state model does not honor the setting of starting soc percent:

```
{'battery_state': {'format': {'floating_point': {'initial': 100.0}},
  'type': 'soc',
  'unit': 'percent'},
 'distance': {'distance_unit': 'miles', 'initial': 0.0},
 'energy_electric': {'energy_unit': 'kilowatt_hours', 'initial': 0.0},
 'time': {'initial': 0.0, 'time_unit': 'minutes'}}
```

When we are initializing the state model we use the default value of 100% and the starting soc percent does not get set properly.

robfitzgerald:
addresses the bug discovered in #241. it appears that:
  1. user-submitted value was correctly validated between 0 and 100
  2. **bug:** the percent value [0,100] was being used directly to scale incoming energy (multiplying it by 100)

so the energy state would always exceed 100%. how this ended up with a state model at 100% is unclear; could be because of `vehicle_ops::as_soc_percent` which clips the value between 0 and 100 again. but i'm not sure.

that said, it is working:

```json
"battery_state": {
        "type": "soc",
        "unit": "percent",
        "format": {
            "floating_point": {
                "initial": 80.0
            }
        },
        "index": 0,
        "name": "battery_state"
    },
```

Closes #241.

### DIFF
--- a/rust/routee-compass-core/src/model/access/default/turn_delays/turn.rs
+++ b/rust/routee-compass-core/src/model/access/default/turn_delays/turn.rs
@@ -1,3 +1,5 @@
+use std::fmt::Display;
+
 use crate::model::access::access_model_error::AccessModelError;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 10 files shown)


nreinicke (approved):
Great find, that was a subtle little bug. Thanks for fixing!

PR #255, opened 2024-09-30
Rjf/map model

### LINKED ISSUE #129: Broaden scope of FrontierModel to include current search tree
Our current trait representing the frontier model is:

```rust
pub trait FrontierModel: Send + Sync {
    fn valid_frontier(
        &self,
        edge: &Edge,
        state: &[StateVar],
        previous_edge: Option<&Edge>,
    ) -> Result<bool, FrontierModelError>;
}
```

optionally providing the previous edge in this scope allows us to restrict access to the next edge in specific applications such as no left turn intersections, but, falls short of allowing us to restrict based on more complex maneuvers that extend beyond 2 edges.

in order to address this, we could replace the previous_edge with the tree:

```rust
pub trait FrontierModel: Send + Sync {
    fn valid_frontier(
        &self,
        edge: &Edge,
        state: &[StateVar],
        tree: &HashMap<VertexId, SearchTreeBranch>,
        direction: &Direction
    ) -> Result<bool, FrontierModelError>;
}
```

Some implications of this change:
  - the direction must also be provided, in the case that the tree is used by a reverse search in a bi-directional algorithm
  - the previous edge can be found by performing a single iteration of backtracking, an _O(1)_ cost

### LINKED ISSUE #244: Map matching + link geometries as first-class modules of a SearchApp
we are currently loading geometries _twice_, once for the map matching input plugin and also for output geometry construction. this is because 1) these assets are not first-class citizens of a SearchApp, 2) the APIs for input plugins and output plugins do not include a &SearchApp in scope. 

we will require some kind of `struct` that provides methods for both of these features, which encompasses all of the existing features:

- vertex/edge-oriented map matching via vertex or edge rtree, constructed based on configuration selection
- edge geometry output plugin processing for building WKT/WKB/GeoJSON outputs

robfitzgerald:
This PR includes the code changes to implement a first class map matching module for RouteE Compass as described in #244. Additionally, updates the frontier model as described in #129.

Status:
  - [x] a map module
    - designed to take references of objects containing geometries, no copy req'd
  - [x] edge rtree in map model
  - [x] storage for geometries in SearchApp
  - [x] wire in map module with SearchApp
  - [x] pass SearchApp to output plugins, read geometries from traversal output plugin
  - [x] pass MapModel to input plugins, or make map matching it's own full step before input processing, and use rtree lookup in map matching phase
  - [x] cleanup/remove old bits
  - [x] pass tests
    - speeds e2e test fails, app.run() returns no rows
  - [x] remove VertexRTree InputPlugin, migrate tests
  - [x] remove EdgeRTree InputPlugin, migrate tests
  - [x] add frontier model road class filtering to map model
  - [x] ~tests for MapInputType::process_origin + process_destination~ agreed to delay on this
  - [x] ~create issue to migrate tests for Traversal OutputPlugin into the MapModel file~ moved assets, agreed to delay
  - [x] fix deserialization of MapInputTypes in the [mapping] section

Closes #244.
Closes #129.


rust/Cargo.toml
```diff
@@ -15,6 +15,7 @@ rayon = "1.7.0"
 geo = { version = "0.28.0", features = ["use-serde"] }
 geo-types = "0.7.12"
 geojson = { version = "0.24.1" }
+proj = "0.27.2"
```
nreinicke, line 18: Looks like this isn't being used currently. Having this in actually causes a build failure on my end (which could probably be resolved) but thinking we should remove it for now until we actually start doing CRS projections.

nreinicke (changes requested):
Nice work, this is looking great. I left a small comment about the proj dependency. 

I also ran this against our open street map example and it failed with the error in the results `request had error: required query field 'origin_vertex' not found`. I tracked this down to the UUID output plugin that expects to find an origin vertex id but in our new defaults, we're using the edge oriented map model and only adding in edge id. 

Eventually, it might be nice to allow the UUID output plugin to support edge uuid mapping in addition to the current vertex only behavior, but at a minimum, I'm wondering if we could remove the UUID output plugin from our default osm config and add in a quick check at the UUID output plugin `process` method to see if we're running an edge based search and return a better error. 

Overall though, everything is running smoothly and it's great to see that we're only loading in our geometries once, great work!

rust/Cargo.toml
robfitzgerald: right. at the outset i was attempting to include the CRS processing that you referred to in our convo yesterday. i'll comment this out and we can come back around to this in the future, but i also had this problem on my m1 and so suspect we may want to make it an optional compass feature.

robfitzgerald:
> I'm wondering if we could remove the UUID output plugin from our default osm config and add in a quick check at the UUID output plugin process method to see if we're running an edge based search and return a better error

right, i thought i already removed it. it's removed now. as for returning a better error, i think instead we need to implement a better UUID plugin that never fails, has two optional input_file arguments for vertex and edge UUIDs, and then simply matches query fields to UUIDs where both field and lookup table exist. for now, it's just dangling there. i'll make an issue. ok, did it: #265.

nreinicke (approved):
That makes sense to me!

PR #256, opened 2024-10-28
Adds in new routee powertrain models from recent training runs

nreinicke:
Adds in a host of new RouteE Powertrain models from a recent training batch.

This brings up a question though: Should we be hosting these binary models in the git repository or would it be cleaner to have them downloaded from a public box folder? The new set of models is about 30MB and if we're updating these frequently, the load on the repository could be high.


nreinicke:
based on an IRL discussion, we should indeed move these files out of the git repo into something else (tbd what that something else is). maybe a git submodule that holds all of the models (for both compass and powertrain)

nreinicke:
@robfitzgerald - I've updated this a bit such that now, only three default models are stored in the git repository (one from each category of ICE, BEV, PHEV). The remaining models are stored in an external box folder and will be downloaded (and cached) when building a compass dataset on the user side. This mirrors the workflow for hosting models from RouteE Powertrain. Eventually I could see us moving to a more robust, versioned approach to this (maybe [DVC](https://dvc.org/) with an s3 remote?) but for now this allows us to reduce the overhead the models impose on the git history.

nreinicke:
These changes has been folded into #268 and so please reference that PR for these changes

PR #259, opened 2024-11-27
Replace 2D interpolation with released `ninterp` crate

kylecarow:


rust/routee-compass-powertrain/src/routee/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -34,27 +33,12 @@ impl PredictionModel for InterpolationSpeedGradeModel {
...
-                        "Could not get first y-value; are y-values empty?".to_string(),
-                    )
-                })?,
-                *interp.y.last().ok_or_else(|| {
-                    TraversalModelError::PredictionModel(
-                        "Could not get last y-value; are y-values empty?".to_string(),
-                    )
-                })?,
+            ninterp::Interpolator::Interp2D(interp) => (
+                // unwraps are okay here, as Vec lengths are checked on instantiation
```
kylecarow, line 37: @robfitzgerald @nreinicke note this comment

kylecarow:
I recommend testing this!! I am pretty unfamiliar with how to run RouteE compass at this point to check this all works alright.

robfitzgerald:
> I recommend testing this!! I am pretty unfamiliar with how to run RouteE compass at this point to check this all works alright.

Pull requests trigger all of our unit tests on the GitHub actions server. But to your point, maybe some unit tests should be part of our acceptance criteria here. We can talk about it next week if you're interested.

nreinicke:
@kylecarow - Thanks for adding this in! I was thinking this would be a good opportunity for us to expand our unit test coverage of the interpolation prediction model by adding a new test that loads both the underlying prediction model and the interpolation model and then makes sure the interpolation model is giving results that are close to the underlying prediction model over the grid of inputs. Do you have any availability to add something like that in?

nreinicke (approved):
Based on an IRL discussion, we decided to check this over our example set and then add an issue for eventually coming back to add a unit test to compare the outputs from our underlying model to the outputs from the interpolation model. 

I've run this over the example set and everything looks good.

PR #260, opened 2024-12-02
Rjf/errors

### LINKED ISSUE #133: simplify error variants
when we started _this journey_ we picked specified taxonomies for traversal model errors, frontier model errors, graph errors and the like, such as `TraversalModelError::MissingIdInTabularCostFunction`. but as we solidify RouteE Compass as a dependency of other downstream projects, those variants are not general enough. what is instead missing is a standardized taxonomy of error types. 

for example, the following variants (incomplete set):
```rust
pub enum TraversalModelError {
    #[error("failure building traversal model from file {0}: {1}")]
    FileReadError(PathBuf, String),
    #[error("failure building traversal model: {0}")]
    BuildError(String),
    #[error("numeric error during calculation: {0}")]
    NumericError(String),
    #[error("index {0} for {1} not found on search state {2:?}")]
    StateVectorIndexOutOfBounds(usize, String, TraversalState),
    #[error("id {0} for id type {1} not found in tabular edge cost function {2}")]
    MissingIdInTabularCostFunction(String, String, String),
    #[error("prediction model failed with error {0}")]
    PredictionModel(String),
}
```

could be reduced to something like
```rust
pub enum TraversalModelError {
    #[error("failure during build for model {1} due to {0}")]
    BuildErrorForModel(String, String),
    #[error("failure during build for service {1} due to {0}")]
    BuildErrorForService(String, String),
    #[error("failure during build for builder {1} due to {0}")]
    BuildErrorForBuilder(String, String),
    #[error("runtime error for model {1} due to {0}")]
    RuntimeErrorForModel(String, String)
}
```

### LINKED ISSUE #183: no transparent errors should bubble to top of app
attempting to run a compass app and i was getting this error:

```
[2024-04-08T22:49:11Z ERROR mep_routee_compass] invalid type: map, expected variant identifier
```

this was due to the fact that my configuration TOML's output policy had double braces instead of single:

```toml

[[response_output_policy]] # should be [...]
type = "file"
```

the error is vague because CompassAppError has a transparent ConfigError (from the config crate):

```rust
    #[error(transparent)]
    ConfigError(#[from] ConfigError),
```

Compass' top-level error types should not be transparent unless they do not require context. this should instead return a BuildError or similar where the developer includes the contextual information:

```rust
    #[error("error building component {0} of type {1}: {2}")]
    BuildError(String, String, String),
```

robfitzgerald:
this is the beginning of a refactor with the intention to address #133. i'm using the time to do housekeeping. creating a draft PR here so we can discuss the changes.

### remove "property" module

before Graph was a struct, it was a trait object, with the idea that we might have different graph representations. we dropped that idea but still Vertex and Edge implementations were living in a separate module called "property" (for Graphs with properties). this PR moves Vertex and Edge into the module with the rest of the graph stuff, and that module has been renamed from road_network to network for simplicity.

### pass EdgeId/VertexId/Direction by reference in Graph methods

we are passing these newtypes by value which i believe triggers a clone. i previously thought the compiler would recognize a built-in under-the-hood but i have heard otherwise on recent searches. this way the signatures match the general idiom of passing immutable values into functions in rust.

for example, a before-and-after:

```rust
// previous
let result = graph.dst_vertex_id(VertexId(0))?;

// new
let result = graph.dst_vertex_id(&VertexId(0))?;
```

### Graph in/out/incident edges methods no longer fail

the in_edges, out_edges, and incident_edges methods previously failed if there wasn't an entry for their argument vertex in the forward or reverse adjacency lists, based on the assumption that we are always working with a strongly-connected graph. while that is sensible, it's not robust, for example if people are working with modified graphs or crowd-sourced datasets. if it truly should produce an error, that can be decided at the calling scope.

### use eprintln! macro instead of println! macro for kdam::Bar

kdam writes the progress bar to stderr, not stdout.

### about errors

i wanted to make things a little more consistent and to reduce fracturing, and so, for a lot of the system, i simplified things to Build, Runtime, and Internal errors, both single-String-parameter enum variants:

```rust
enum MyError {
  #[error("failure building my model due to {0}")]
  BuildFailure(String),
  #[error("failure while running my model due to {0}"]
  RuntimeFailure(String),
  #[error("{0}")]
  InternalError(String)
}
```

so:
- if we had a bunch of different variants that seemed to fit into these categories, i removed those variants
- other modules seemed to benefit from being more specific, and that felt appropriate lower-down in the module stack, like the road network operations
- i revised the composition between error types so it's cleaner
- more information is communicated between nested errors where appropriate (see SearchError)
- removed transparent error variants (#183)
- leveraged the macro-generated "From" implementations for easy error type propagation 
  - for example, in EdgeTraversal, a bunch of errors come from different models, but the compiler calls the SearchError::from method when `?` is called on those results

with these changes, CompassAppErrors should now have the right amount of context for most situations. 

Closes #133.
Closes #183.


nreinicke (approved):
This looks great. I really appreciate the additional error context and improved layering that you've introduced here.

PR #267, opened 2024-12-12
Rjf/bug route on empty tree

robfitzgerald:
This little PR modifies the vertex backtrack function so that, when passed an empty tree, it terminates early with an empty vector. An empty tree is a valid search result for the case when the origin location and destination location match or are incident edges.


nreinicke (approved):
Yep this makes sense

PR #268, opened 2024-12-12
Update Examples

nreinicke:
Builds off of #238 to update our examples. Now, we have three main examples:
 - Open Street Maps Example: The goto beginning example that shows how to build a dataset and run the application
 - Different Powertrain Example: Dives a bit deeper into the differences between running different powertrains
 - Time Energy Tradeoff Example: Dives deeper into comparing time and energy tradeoffs
 
I also moved the examples to standard python scripts which should make them much easier to manage with version control / linting / formatting. 

When building the docs, we can now call `python docs/examples/_convert_examples_to_notbooks.py` which converts all the examples to notebooks prior to building the docs since jupyter book expects them.

Closes #238


robfitzgerald (approved):
nice setup, should be easy to add examples and avoid stale notebook state. i think the only drawback with that is that the notebook output cells are not rendered, so, that makes reading the examples on the web somewhat ineffective. i'm happy to approve though and maybe we try and figure out how to render those cells on release.

thanks!

nreinicke:
> i think the only drawback with that is that the notebook output cells are not rendered, so, that makes reading the examples on the web somewhat ineffective.

Yeah that's true that they won't be rendered from within GitHub but if everything works as expected, the outputs will be rendered in the html documentation once it gets built. Would that address your concern?

robfitzgerald:
Oh it does? Ya that's actually purrrfect

PR #269, opened 2024-12-13
Update dependencies and unit test

### LINKED ISSUE #266: Unit test to validate base Powertrain model against interpolation model
In #259 we accepted a revised linear interpolation library. Linear interpolation is leveraged by the interpolated powertrain model in order to speed up inference. Interpolation approximates the underlying Powertrain model. As of now we don't have any unit test confirming that the outputs of the original model are equivalent with some acceptable error.

nreinicke:
A small PR to pick off a couple of issues:
 - Closes #252 by updating most dependencies to the latest version. The exception to this is the wkb crate as the move from 0.7.1 to 0.8 introduced non-trivial breaking changes and so I opened a new issue #270 for that one bump
 - Closes #266 by making or existing unit test for the interpolation model more robust to sweep over a range of inputs and make sure both the interpolation model and the underlying smartcore model are producing similar results.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,25 +10,25 @@ members = [
 
 [workspace.dependencies]
 serde = { version = "1.0.209", features = ["derive"] }
-serde_json = { version = "1.0.127", features = ["preserve_order"] }
-rayon = "1.7.0"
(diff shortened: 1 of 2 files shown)


robfitzgerald (approved):
sweet, thank you for the added test as well!

PR #271, opened 2024-12-19
log to_string_pretty error in debug instead of fail

robfitzgerald:
in the process of using routee incorrectly (sigh) i realized that the debug plugin could fail if the value passed to it was not JSON deserializable. instead of raising that serde_json::Error, i have the debug plugin write a pretty-printed error message in the debug log instead.

not tested, but, since debug is not in the critical path of Compass, we could probably accept this.


nreinicke (approved):
This makes sense to me!

PR #278, opened 2025-01-23
Upgrade ninterp to 0.2.0

kylecarow:
Hey dudes

I just released ninterp v0.2.0, which has a minor syntax change to make the internal structs totally private. This changes the syntax as shown below:

Before:
```rust
let interp = Interpolator::Interp2D(
    Interp2D::new(
        ...
    )?
);
```

After:
```rust
let interp = Interpolator::new_2d(
    ...
)?;
```

Since the `Interp1D`/`Interp2D`/etc. structs are private, matching for N > 0 should be done with `..`:
```rust
match interp {
    Interp0D(value) => {},
    Interp1D(..) => {},
    Interp2D(..) => {},
    Interp3D(..) => {},
    InterpND(..) => {},
}
```

This is in response to some occurrences in FASTSim when we accidentally called e.g. `Interp2D::interpolate`, rather than the correct `Interpolator::interpolate`. This was sneakily skipping the checks on data ranges (to be handled according to `Extrapolate` selection).

I also think the new syntax is a little tidier and clearer.

This shouldn't affect routee-compass at all, but I figured having the latest syntax wouldn't hurt.

### DIFF
--- a/rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
+++ b/rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
@@ -2,6 +2,7 @@ use super::utils::linspace;
 use crate::model::prediction::{
     load_prediction_model, model_type::ModelType, prediction_model::PredictionModel,
 };
+use ninterp::*;
 use routee_compass_core::{
     model::traversal::TraversalModelError,
     model::unit::{
@@ -11,7 +12,7 @@ use routee_compass_core::{
 use std::path::Path;
 
 pub struct InterpolationSpeedGradeModel {
-    interpolator: ninterp::Interpolator,
+    interpolator: Interpolator,
     speed_unit: SpeedUnit,
     grade_unit: GradeUnit,
     energy_rate_unit: EnergyRateUnit,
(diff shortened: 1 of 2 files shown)


rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -28,37 +29,34 @@ impl PredictionModel for InterpolationSpeedGradeModel {
...
-                })?,
-
-            ),
-            _ => {
-                return Err(TraversalModelError::TraversalModelFailure(
-                    "Only 2-D interpolators are currently supported".to_string(),
-                ))
-            }
-        };
+        // NOTE: `x().unwrap()` and `y().unwrap()` are okay, because they never fail for `Interpolator::Interp2D`
```
kylecarow, line 39: @nreinicke @robfitzgerald Make sure to take note of this. Of course we could remove the unwrap, but these methods never fail for 2D interpolators, so it would just be adding unnecessary handling.

robfitzgerald:
Yes to handling the error case instead of unwrapping. I'm away from my laptop, hence brevity. Thank you!

nreinicke:
Nice! Thanks for updating this! And agreed that we should integrate it without using unwrap and actually handing the Error variant of the result. Even if the function can't fail now we want our code style to be consistent and if the logic changes in the future and we forget to update our code we'll get a panic.

kylecarow:
Closing in favor of #308

PR #281, opened 2025-02-05
Rjf/279 simplify unit configurations and conversions

### LINKED ISSUE #279: Update traversal model unit representation
Our current traversal model unit representation could be a little bit clearer. For example, we might have a config that looks like this:

```toml
[traversal]
type = "energy_model"
time_model_speed_unit = "kilometers_per_hour"
grade_table_input_file = "edges-grade-enumerated.txt.gz"
grade_table_grade_unit = "decimal"

time_unit = "minutes"
distance_unit = "miles"

[traversal.time_model]
type = "speed_table"
speed_table_input_file = "edges-posted-speed-enumerated.txt.gz"

speed_unit = "kilometers_per_hour"
distance_unit = "miles"
time_unit = "minutes"
```

Looking at the traversal.time_model section, it's unclear what the units represent. For example, the `speed_unit` field is used to indicate what units will be found in the `speed_table_input_file` but the `distance_unit` is used as the unit that the time model outputs. We also have the key `time_model_speed_unit` which could be mistakenly set differently from the `speed_unit` key with no error thrown.

Let's think about how to make these more clear for the user even if that's just updating their name so that `speed_unit` might `speed_table_speed_unit` and `time_unit` might become `output_time_unit`. 

We might also be able to get the time model speed units directly from the model rather than making the user set it.

robfitzgerald:
This PR begins the work described in [this milestone](https://github.com/NREL/routee-compass/milestone/9) by introducing a `Convert` trait for units. changes in this PR are due to that change and the discovery + design task for the scope of this milestone. 

along the way:
  - SpeedUnit is now a tuple newtype struct of Distance and Time units
    - conversion of SpeedUnits leans on the underlying distance + time conversion factors
  - EnergyRateUnit is now an enum, each variant a tuple newtype struct of Energy and Distance units
    - comes in distance/energy and energy/distance variants to support convention (mpg) vs our preference during routing (gpm)
  - `Convert` methods take a [`Cow<U>`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) for the given unit type. this allows us to avoid unnecessary copying when the conversion is between two of the same unit type (such as kwh -> kwh)
  - FromStr, Display, Serialize, and Deserialize ops expanded to allow for some more human-legible configurations, like using "kph" or "kilometers/hour" for a speed unit value, or "km" for kilometer, or "kWh" for kilowatt hours
  - added a few tests for these new capabilities (deserialize + convert)

Closes #279.


robfitzgerald:
i kept running into confusion as to whether an energy rate unit should follow the well-known convention of distance/energy unit, or if we should impose energy/distance unit, since that is more efficient for routing. but now i'm thinking, we should support _both_, and call me crazy, but i've started making that change:

```rust
pub enum EnergyRateUnit {
    DistancePerEnergy(DistanceUnit, EnergyUnit),
    EnergyPerDistance(EnergyUnit, DistanceUnit),
}
```

internally we can convert things to `EnergyRateUnit::EnergyPerDistance(KilowattHours, Meters)`. this means i need to add a `Convert` implementation for `EnergyRateUnit`.

robfitzgerald:
@nreinicke ok, now ready for review. also added another issue for volume in energy units which would further clean things up (see #293, maybe a good first issue?)

nreinicke:
Sorry that I'm just getting around to reviewing this. I started by just running this against the main branch and I'm noticing that this changes the magnitude of energy consumption for the Toyota Camry. For example, in our 03_time_energy_tradeoff_example.py script, we get different results:

### This PR
image

### Main branch
image

The total energy consumed for the energy optimal case drops down by 20% in this PR. I'm not sure where this discrepancy is coming from but given how big it is we probably want to convince ourselves that we're not doing anything wrong here (or in the main branch).

rust/routee-compass-powertrain/src/model/energy_traversal_model.rs
```diff
@@ -60,18 +61,31 @@ impl TraversalModel for EnergyTraversalModel {
...
                 .energy_model_service
                 .time_model_speed_unit
                 .associated_distance_unit(),
-        );
-        let speed = Speed::from((distance_in_time_model_unit, time_delta));
+        )?;
+        let speed_tuple = Speed::from_distance_and_time(
+            (
+                &distance_in_energy_model_unit,
+                &self.energy_model_service.distance_unit,
```
nreinicke, line 75: I think I found the issue. When creating the speed here, we're telling it it's in the `energy_model_service.distance` unit (miles) but that distance has actually been converted to `energy_model_service.time_model_speed_unit.associated_distance_unit()` (kilometers). If I change like 75 to `&self.energy_model_service.time_model_speed_unit.associated_distance_unit(),`, the outputs are matching what we see in the main branch.

I think this highlights one of the main issues with the unit system that we should address which is that it's easy to mix up units if they're not explicitly paired together. I wonder if we can put together an abstraction such that we're required to pass around units with values.

robfitzgerald: Nice catch. I agree if we could do that things would make more sense, but then we may greatly increase allocations by shipping around unit enum instances with every arithmetic operation. The current design is trying to avoid all of that cloning. I would love to see other ideas.

nreinicke: Yeah that's a good point. I haven't spent enough time thinking about this to propose a good solution but hopefully we can eventually work towards something like this.

robfitzgerald:
@nreinicke thank you for diving into the results and finding my error! the change you proposed has been added.

rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
```diff
@@ -104,23 +111,35 @@ impl TraversalModel for SpeedTraversalModel {
         if distance == Distance::ZERO {
             return Ok(());
         }
+        // <<<<<<< HEAD
```
nreinicke, line 114: Looks like there are some relics of the merge left over here (unless you had intended to leave them in?)

robfitzgerald: SAY WHAAAAAT????????? 😱

nreinicke (approved):
Beautiful, this looks great, thanks for making these changes!

PR #290, opened 2025-02-17
pipeline phases, requests kwds

### LINKED ISSUE #288: split out generate_dataset operations
i generated a dataset which completed most steps but failed on the last due to an SSL error (proxy settings):

the [code to run just the model import](https://github.com/NREL/routee-compass/blob/main/python/nrel/routee/compass/io/generate_dataset.py#L214) is not a separate function, so i can't simply re-run that step. nor can i specify which steps i want to run as an argument for the generate_compass_dataset function. a refactor to separate phases out and make it possible to automatically find the next step to run or name it explicitly at generate_compass_dataset would make this easier:

```python
generate_compass_dataset(g, "output_dir", phases=None) # default behavior

generate_compass_dataset(g, "output_dir", phases=["grade", "powertrain"])  # re-run a few phases

generate_compass_dataset(g, "output_dir", phases=["powertrain"])  # re-run just that last one
```

### LINKED ISSUE #289: proxy-friendly generate_dataset workflow
generate_compass_dataset will fail when behind a proxy for a typical user. our code should support proxy configuration by passing it along to methods making the box.com GET calls to grab powertrain models.

robfitzgerald:
this PR adds features to dataset generation to allow users to:
  - run specific phases of the generation pipeline
  - pass keyword arguments to the requests.get call for credentials/proxy/etc reasons

i would like some feedback on the name of the "phases" argument. it could be "steps", or something else, if that's clearer.

Closes #288.
Closes #289.

### DIFF
--- a/docs/examples/01_open_street_maps_example.py
+++ b/docs/examples/01_open_street_maps_example.py
@@ -44,14 +44,16 @@ g = ox.graph_from_place("Denver, Colorado, USA", network_type="drive")
 Now, we call the `generate_compass_dataset` function which will convert the osmnx graph into files that are compatible with RouteE Compass.
 
 ```{note}
-In order to get the most accurate energy results from the routee-powertrain vehicle models, it's important to include road grade information since it plays a large factor in vehicle energy consumption (`add_grade=True`)
+In order to get the most accurate energy results from the routee-powertrain vehicle models, it's important to include road grade information since it plays a large factor in vehicle energy consumption.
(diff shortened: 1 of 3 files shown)


nreinicke (commented):
Nice, this is much cleaner than before. Would it make sense to make phase a simple enum to catch a user accidentally misspelling one of the phases?

robfitzgerald:
I forgot that this was my PR 😊  I can still enum it!

nreinicke:
Haha no worries! I just saw that the remaining fix was a couple simple mypy fixes and so I just went in and did it since they weren't controversial 😄

nreinicke (approved):
LGTM! 👍

PR #292, opened 2025-02-19
Rjf/string road class

### LINKED ISSUE #291: road class frontier model should support String road class types
currently the road class model only allows classes that are `u8` values. these could be `String` values instead, without perceivable loss of performance, since lookup in a HashSet<u8> and HashSet<String> should be the same. at the same time, we could support [OSM Highway keys](https://wiki.openstreetmap.org/wiki/Key:highway) without requiring the user to map them to an arbitrary number.

robfitzgerald:
Modify RoadClassFrontierModel so it supports arbitrary string road classes. Supports previous functionality but without a RoadClassParser, since numeric road class values (like TomTom routing class values) are also valid strings. Should be no loss of performace since HashSet.contains() for u8 or String values are both _O(1)_.

This allows us to directly support OSM highway tags as road class filters without forcing the user to enumerate them.

Closes #291.


nreinicke (approved):
Yeah I like this, it's much more flexible.

Issue #257, opened 2024-10-29
Allow vehicle to have max speed


nreinicke:
Our current time traversal model uses a speed lookup table to determine the vehicle's speed on any given road link. But, in some instances of "eco-driving" the vehicle might be better suited in driving a speed that is lower than the speed reported by the road network. For example, on an interstate with free flow speeds at or above 80 mph, a battery electric vehicle might save significant energy by reducing its speed by some amount. To that end, we could create a new time traversal model that would permit the vehicle to drive at some percentage of the road network speed. This should be user configurable and smart such a vehicle is not permitted to drive at speeds that are unrealistic (think 20mph on an 80mph interstate).

PR #295, opened 2025-03-19
Ndr/257 max speed

nreinicke:
Implements #257 by adding a new `speed_limit` to the speed traversal model such that a vehicle cannot exceed this speed when traversing.

### DIFF
--- a/rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
+++ b/rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
@@ -13,25 +13,26 @@ use std::sync::Arc;
 pub struct SpeedTraversalModel {
     engine: Arc<SpeedTraversalEngine>,
 
-    max_speed: Option<Speed>, // max speed is converted to the speed unit of the engine
+    speed_limit: Option<(Speed, SpeedUnit)>,
 }
 
 impl SpeedTraversalModel {
     pub fn new(
         engine: Arc<SpeedTraversalEngine>,
-        max_speed: Option<(Speed, SpeedUnit)>,
+        speed_limit: Option<(Speed, SpeedUnit)>,
     ) -> SpeedTraversalModel {
-        if let Some((max_speed, max_speed_unit)) = max_speed {
+        if let Some((max_speed, max_speed_unit)) = speed_limit {
             let converted_speed = max_speed_unit.convert(&max_speed, &engine.speed_unit);
+            let converted_speed_unit = engine.speed_unit;
             return SpeedTraversalModel {
                 engine,
-                max_speed: Some(converted_speed),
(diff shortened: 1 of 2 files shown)


rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
```diff
@@ -32,6 +49,18 @@ impl TraversalModel for SpeedTraversalModel {
         let (_, edge, _) = trajectory;
         let distance = BASE_DISTANCE_UNIT.convert(&edge.distance, &self.engine.distance_unit);
         let speed = get_speed(&self.engine.speed_table, edge.edge_id)?;
+
+        let speed = match self.speed_limit {
+            Some((speed_limit, _speed_unit)) => {
```
robfitzgerald, line 54: Since the user can provide a speed unit, we have to first call convert to make sure we are correctly comparing speeds in the same unit, right?

rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
```diff
@@ -263,4 +297,79 @@ mod tests {
         // approx_eq(result.updated_state[1].into(), expected, 0.001);
         approx_eq(state[1].into(), expected, 0.001);
     }
+
+    #[test]
```
robfitzgerald, line 301: If I'm right about my other comment, then maybe we could also add testing where the limit unit is different

robfitzgerald (changes requested):
Dig it, sensible feature. I have a request for handling the configured speed unit value. Also I'm on my phone so I can't see clearly for myself, but can you confirm that this can be set at query time?

rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
nreinicke: Yes, the idea is to convert it in the `new` method to match the units of the speed model such that it only has to be converted once.

rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
nreinicke: Yeah that's a good idea, I'll add that

nreinicke:
> can you confirm that this can be set at query time?

Yes, the service takes the keys `speed_limit` and `speed_limit_units` from the user query.

rust/routee-compass-core/src/model/traversal/default/speed_traversal_model.rs
robfitzgerald: Oh ok, good point. Mind leaving a comment here about that? I'm on board.

nreinicke: Yes, good idea, just added

PR #298, opened 2025-03-25
Fix bug in generate dataset

nreinicke:
Fixes a small regression from #290 where logic for writing the graph was inverted


robfitzgerald (approved):
Ooooop! Thanks Nick ☺️

PR #303, opened 2025-04-16
Rjf/dependencies bump

robfitzgerald:
this PR bumps all dependencies to latest except ninterp, which we should reboot discussions with @kylecarow about.

a few api changes to dependencies:

##### bincode (major version)
  - changes to signatures
  - we are using a "legacy" config for bincode 1.0 compatibility. when RouteE Powertrain models move to bincode 2.0.0 we should modify this
  - we should even _parameterize this_ so that we support v1 and v2 bincode files

##### wkb

complete rewrite of the wkb library between 0.7 and 0.8 has new api and hopefully is more reliable (it has produced invalid WKBs in bambam)


nreinicke (approved):

kylecarow:
Regarding a ninterp version bump, see new PR #308

PR #310, opened 2025-05-01
Rjf/unit deserializing

### LINKED ISSUE #309: inconsistent use of FromStr / Deserialize of some units
The EnergyRateUnit deserializer is customized so that it expects strings of the format "<energy_unit>/<distance_unit>" or "<distance_unit>/<energy_unit>". under-the-hood, it uses FromStr<EnergyUnit> and FromStr<DistanceUnit>. however, EnergyUnit and DistanceUnit use a default Deserialize method. the strings they support are not the same, so:

```toml
# valid EnergyUnit deserializing uses the serde::Deserialize macro with `rename_all = "snake_case"`
energy_unit_01 = "kilowatt_hours"
# valid EnergyRateUnit deserializing uses FromStr<EnergyUnit> instead:
energy_rate_unit_01 = "kilowatthour/miles"
energy_rate_unit_02 = "kilowathours/miles" # <-- also, typo here
energy_rate_unit_03 = "kwh/miles"
```

these should be unified, via leveraging FromStr implementations in a custom deserialize implementation in DistanceUnit + EnergyUnit.

robfitzgerald:
This PR makes sure FromStr and Deserialize behaviors match for the unit enums according to the issue described in #309. In order to unify them, the enums now use the [`try_from` container attribute](https://serde.rs/container-attrs.html#try_from) which point to the FromStr implementation via a TryFrom<String> implementation. 

Along the way,
  - some missing abbreviations/pluralizations for unit deserializing were added
  - documentation on unit abbreviations was added to the jupyterbook docs

Closes #309.

### DIFF
--- a/rust/routee-compass-core/src/model/unit/energy_unit.rs
+++ b/rust/routee-compass-core/src/model/unit/energy_unit.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Copy)]
-#[serde(rename_all = "snake_case")]
+#[serde(rename_all = "snake_case", try_from = "String")]
 pub enum EnergyUnit {
     /// US Gallons of gasoline fuel
     GallonsGasoline,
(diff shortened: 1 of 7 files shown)


rust/routee-compass-core/src/model/unit/energy_unit.rs
```diff
@@ -84,13 +84,26 @@ impl FromStr for EnergyUnit {
...
-            "liters gasoline" => Ok(E::LitersGasoline),
-            "liters diesel" => Ok(E::LitersDiesel),
+        match s
+            .trim()
+            .to_lowercase()
+            .replace("_", " ")
+            .replace(" ", "")
+            .as_str()
+        {
+            "gallonsgasoline" | "gas" => Ok(E::GallonsGasoline),
```
nreinicke, line 94: I'm wondering if it might be too ambiguous to allow "gas" to translate to gallons but not to liters? Perhaps it's better to require that you specify gallons or liters so it's not unclear about which one it maps to? Same with diesel.

nreinicke (commented):
This looks good, thanks for making this more clear. I left a small comment about what "gas" or "diesel" default to but otherwise looks good.

rust/routee-compass-core/src/model/unit/energy_unit.rs
robfitzgerald: you're right that this is definitely a US-centric take on a common abbreviation. i'll do as you suggest and remove "gas" and "diesel" aliases.

nreinicke (approved):

PR #313, opened 2025-05-02
test input fix

robfitzgerald:
small fix to make test pass


robfitzgerald:
@kylecarow here's a suggested change to your pull request to fix that broken test so we can accept your work

PR #316, opened 2025-05-07
Utilize inbuilt ninterp clamp

kylecarow:
@robfitzgerald @nreinicke My apologies for catching this improvement moments after merging my last PR. Hate it when that happens!

Anyway, I recognized that routee-compass should just be using the `Extrapolate::Clamp` setting provided in ninterp, rather than clamping manually before interpolation. This saves us some lines in this repo, and utilizes functionality already present in ninterp. This should be a very fast review.

Hopefully squeaking this PR in right after the last one is unintrusive enough!

### DIFF
--- a/rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
+++ b/rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
@@ -32,34 +32,9 @@ impl PredictionModel for InterpolationSpeedGradeModel {
         speed_unit.convert(&mut speed_converted, &self.speed_unit)?;
         grade_unit.convert(&mut grade_converted, &self.grade_unit)?;
 
-        // snap incoming speed and grade to the grid
-        let min_speed = *self.interpolator.data.grid.get(0).map(|s| s.first()).flatten().ok_or_else(|| {
-            TraversalModelError::TraversalModelFailure(
-                "Could not get first speed value from powertrain interpolation result; are x-values empty?".to_string(),
-            )
-        })?;
-        let max_speed = *self.interpolator.data.grid.get(0).map(|s| s.last()).flatten().ok_or_else(|| {
-            TraversalModelError::TraversalModelFailure(
-                "Could not get last speed value from powertrain interpolation result; are x-values empty?".to_string(),
-            )
-        })?;
-        let min_grade = *self.interpolator.data.grid.get(1).map(|g| g.first()).flatten().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
```diff
@@ -32,34 +32,9 @@ impl PredictionModel for InterpolationSpeedGradeModel {
         speed_unit.convert(&mut speed_converted, &self.speed_unit)?;
         grade_unit.convert(&mut grade_converted, &self.grade_unit)?;
 
-        // snap incoming speed and grade to the grid
-        let min_speed = *self.interpolator.data.grid.get(0).map(|s| s.first()).flatten().ok_or_else(|| {
```
robfitzgerald, line 36: is there something in ninterp that confirms the grid is non-empty during initialization or before we start using it?

robfitzgerald (commented):
thanks Kyle! i have a question about the removal of those boundary checks, but assuming that checks out, this looks good to me.

rust/routee-compass-powertrain/src/model/prediction/interpolation/interpolation_speed_grade_model.rs
kylecarow: Yep!

During initialization (i.e. calling the `new` method like we do in this file) interpolators run the `InterpData<_, _>::validate` method, which does these checks:

https://github.com/NREL/ninterp/blob/fe9cb8e3221f751785ff4bc7bbfbd918708edb3f/src/interpolator/data.rs#L64-L81
```rust
pub fn validate(&self) -> Result<(), ValidateError> {
    for i in 0..N {
        let i_grid_len = self.grid[i].len();
        // Check that each grid dimension has elements
        if i_grid_len == 0 {   /// <-------------------------------- RELEVANT LINES HERE
            return Err(ValidateError::EmptyGrid(i));
        }
        // Check that grid points are monotonically increasing
        if !self.grid[i].windows(2).into_iter().all(|w| w[0] <= w[1]) {
            return Err(ValidateError::Monotonicity(i));
        }
        // Check that grid and values are compatible shapes
        if i_grid_len != self.values.shape()[i] {
            return Err(ValidateError::IncompatibleShapes(i));
        }
    }
    Ok(())
}
```

This means that if the coordinate grid arrays were empty, we would get a Err result from the constructor.

This guarantee of a non-empty grid is only void if you manually edit the interpolator data without calling `validate` afterwards.

nreinicke (approved):
Yep this looks good, thanks!

PR #320, opened 2025-05-15
more complete serialization of state and cost at each route/tree branch

robfitzgerald:
added some improvements here to route + tree serialization so they now serialize using the state model and cost model to add those sections to each link in a route or tree, like this:

```json
        "properties": {
          "traversal": {
            "edge_id": 7563,
            "access_cost": 0.008333333384705321,
            "traversal_cost": 0.05487432604079934,
            "result_state": [
              0.2613151025581488,
              0.10974865208159863,
              0.6234684593965638,
              0.06473915841810438,
              0.01360034135298995,
              99.89210140263648,
              48.3,
              0.0,
              0.05489683033149421
            ]
          },
          "state": {
            "trip_time": 0.6234684593965638,
            "trip_energy": 0.06473915841810438,
            "trip_soc": 99.89210140263648,
            "trip_distance": 0.2613151025581488
          },
          "cost": {
            "total_cost": 0.4906643208840419,
            "trip_energy": 0.007768699010172525,
            "trip_distance": 0.17116139217558748,
            "trip_time": 0.3117342296982819
          }
```

this allows me to develop plots that include tree and route.

### DIFF
--- a/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
+++ b/rust/routee-compass-core/src/algorithm/search/a_star/a_star_algorithm.rs
@@ -141,38 +141,38 @@ pub fn run_vertex_oriented(
         solution.len()
     );
 
-    #[cfg(debug_assertions)]
-    {
-        use std::io::Write;
-        use std::path::PathBuf;
+    // #[cfg(debug_assertions)]
+    // {
+    //     use std::io::Write;
+    //     use std::path::PathBuf;
 
-        log::debug!("Building flamegraph for search memory usage..");
-        let mut flamegraph = allocative::FlameGraphBuilder::default();
-        flamegraph.visit_root(&costs);
-        flamegraph.visit_root(&traversal_costs);
-        flamegraph.visit_root(&solution);
-        let output = flamegraph.finish_and_write_flame_graph();
+    //     log::debug!("Building flamegraph for search memory usage..");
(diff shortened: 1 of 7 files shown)


robfitzgerald:
@nreinicke i'm using this to look into why things are radically different with energy vs time routes. and yes, there's something wrong with the energy search. i'm going to spare an interpretation for now and just keep investigating as to why.

_also, it looks like somewhere between route generation and plotting, the route order gets mixed up, so that the first and last linestrings in the feature collection are not the origin and destination (see the markers). this may just be something i'm doing wrong in python, but still, it's counter-intuitive if that's the case._

energy_0

energy_50

energy_84

energy_99

robfitzgerald:
whereas the ICE vehicle with energy weight 100% looks like a normal:

ice_energy_99

providing these plots for reference, since they put another perspective on what's happening:

image

image

i'll review changes in my PR that could affect the cost estimation for the a* heuristic in the BEV case.

robfitzgerald:
🤦‍♂️ forgot to incorporate the ideal energy rate during "estimate traversal" for BEV and PHEV 🤦‍♀️

nreinicke:
> 🤦‍♂️ forgot to incorporate the ideal energy rate during "estimate traversal" for BEV and PHEV 🤦‍♀️

So glad you found this!! It makes sense once you start visualizing things in this way

PR #324, opened 2025-05-23
Implemented `VolumeUnit` and extended `EnergyUnit` variants and tests

### LINKED ISSUE #293: EnergyUnit expansion + volume unit for liquid EnergyUnits
across different countries and sectors the volume unit for liquid fuel type varies. we began weaving this into Compass `EnergyUnit`s:

```rust
pub enum EnergyUnit {
    /// US Gallons of gasoline fuel
    GallonsGasoline,
    /// US Gallons of diesel fuel
    GallonsDiesel,
    /// electric fuel
    KilowattHours,
    /// SI liters of gasoline fuel
    LitersGasoline,
    /// SI liters of diesel fuel
    LitersDiesel,
}
```

but there are still a few major categories missing, such as Joules, UK gallons, and miles per gallon equivalent. to break down these variations, we can first create a `VolumeUnit` type:

```rust
pub enum VolumeUnit {
  GallonsUs,
  GallonsUk,
  Liters
}
```

with that, we can parameterize liquid fuel types and expand to allow additional common categories:

```rust
pub enum EnergyUnit {
  KilowattHours,
  Gasoline(VolumeUnit),
  Diesel(VolumeUnit),
  GallonsGasolineEquivalent,
  GallonsDieselEquivalent,
}
```

yamilbknsu:
The `VolumeUnit` enum was implemented in order to allow energy units defined in terms of volume of fuel (such as Gallons of Gasoline vs Liters of Gasoline)

The `EnergyUnit::convert` implementation aims to minimize the number of arithmetic operations required for each call. An alternative that transform to a base first (e.g. `to_kwh`) would be much less verbose but have larger performance impact.

closes #293

### DIFF
--- a/rust/routee-compass-core/src/model/unit/energy_unit.rs
+++ b/rust/routee-compass-core/src/model/unit/energy_unit.rs
@@ -26,6 +26,14 @@ fn get_volume_factor(from: &VolumeUnit, to: &VolumeUnit) -> Result<f64, UnitErro
     Ok(volume_value.as_f64())
 }
 
+impl EnergyUnit {
+    #[allow(non_upper_case_globals)]
+    const kWh: Self = Self::KilowattHours;
+    const KJ: Self = Self::KiloJoules;
+    const GDE: Self = Self::GallonsDieselEquivalent;
+    const GGE: Self = Self::GallonsGasolineEquivalent;
+}
+
 #[allow(non_upper_case_globals)]
 impl Convert<Energy> for EnergyUnit {
     fn convert(&self, value: &mut std::borrow::Cow<Energy>, to: &Self) -> Result<(), UnitError> {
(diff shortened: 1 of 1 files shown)


rust/routee-compass-core/src/model/unit/energy_unit.rs
```diff
@@ -1,66 +1,165 @@
...
-            (S::LitersGasoline, S::GallonsGasolineEquivalent) => Some(0.264172),
-            (S::LitersDiesel, S::GallonsGasolineEquivalent) => Some(3.78541 * 1.14),
-            (S::GallonsGasolineEquivalent, S::GallonsGasoline) => None,
-            (S::GallonsGasolineEquivalent, S::GallonsDiesel) => Some(0.8771929825),
-            (S::GallonsGasolineEquivalent, S::KilowattHours) => Some(33.3333333333),
-            (S::GallonsGasolineEquivalent, S::LitersGasoline) => todo!(),
-            (S::GallonsGasolineEquivalent, S::LitersDiesel) => todo!(),
-            (S::GallonsGasolineEquivalent, S::GallonsGasolineEquivalent) => None,
+            // Variants that do not need transformation
+            (x, y) if x == y => None,
```
robfitzgerald, line 46: let's remove the default line below and replace this one with 

```rust
(EU::KilowattHours, EU::KilowattHours) => None,
(EU::GallonsGasolineEquivalent, EU::GallonsGasolineEquivalent) => None,
(EU::GallonsDieselEquivalent, EU::GallonsDieselEquivalent) => None,
(EU::KiloJoules, EU::KiloJoules) => None,
(EU::BTU, EU::BTU) => None,
```

detail: if we need to add new variants of Volume or Energy in the future, we won't get a compiler warning here that we forgot to handle them because of the default arm.

rust/routee-compass-core/src/model/unit/energy_unit.rs
```diff
@@ -1,66 +1,165 @@
...
+            (EU::Gasoline(V::GallonsUs), EU::GallonsGasolineEquivalent)
```
robfitzgerald, line 47: we could probably reduce a lot of these match cases to a single line, for readability, using aliases like so:

```rust
impl EnergyUnit {
    #[allow(non_upper_case_globals)]
    const kwH: Self = Self::KilowattHours;
    const GDE: Self = Self::GallonsDieselEquivalent;
    const GGE: Self = Self::GallonsGasolineEquivalent;
}
```

robfitzgerald (changes requested):
smooth! thanks for the contribution. i think from your implementation we shouldn't break any existing configurations, in particular that deserialize EnergyUnit instances from TOML config files. i have a small request for a change to the EnergyUnit Convert match clause and an optional comment to improve readability. thanks Yamil!

yamilbknsu:
Thanks @robfitzgerald! I incorporated these comments. Let me know if it's looking better

robfitzgerald (approved):
dig it, thanks!

PR #325, opened 2025-05-23
Ndr/135 n prediction features

### LINKED ISSUE #135: n-dimensional interpolation models

nreinicke:
Closes #135 by updating our vehicle models to accept N input features rather than hard coding them as speed and grade model. Big thanks to @kylecarow for unlocking this with the ninterp crate. 

Makes some additional changes along the way:
 - removes the onnx feature; I would like to revisit this in the future but since we're not using it the code was a bit out of date with the latest ort runtime and would take some effort to build it back up.
 - removes the FloatCache policy; this was an old attempt to cache energy predictions but is not being used.
 - updates to the `tomlkit` library that support the TOML 1.0 specification. Our existing `toml` library did not support the latest standard and `tomlkit` is recommended by the [official python docs](https://docs.python.org/3/library/tomllib.html) for our use case (editing/writing existing toml files).
 - updates the default VehicleCostRate to be Raw rather than Zero. When it defaulted to zero it was causing problems if you did not define a vehicle rate for one of the state variables, rendering it zero whereas I think the Raw default is safer since if a user doesn't add a vehicle cost, it will still pass the variable through. 
 - Sets the input feature internal unit to be required rather than optional; I was confused by the concept of an optional unit in the input feature and so I'm proposing that we update it to just use the baseunit if the units truly don't mater to the traversal model rather than making it optional.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "maturin"
 
 [project]
 name = "nrel.routee.compass"
-version = "0.11.1"
+version = "0.11.2"
 description = "An eco-routing tool build upon RouteE-Powertrain"
 readme = "README.md"
 documentation = "nrel.github.io/routee-compass"
--- a/rust/routee-compass-core/Cargo.toml
+++ b/rust/routee-compass-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "routee-compass-core"
-version = "0.11.1"
+version = "0.11.2"
 edition = "2021"
 exclude = ["test", "**/.DS_Store", "target/"]
 readme = "README.md"
--- a/rust/routee-compass-core/README.md
+++ b/rust/routee-compass-core/README.md
@@ -12,7 +12,7 @@ To install as a library in Rust, add routee-compass-core to your Cargo.toml file
 
 ```toml
 [dependencies]
-routee-compass-core = { version = "0.11.1" }
+routee-compass-core = { version = "0.11.2" }
 ```
 
 Please see the [documentation](https://docs.rs/routee-compass-core/latest/routee_compass_core/) for usage.
--- a/rust/routee-compass-macros/Cargo.toml
+++ b/rust/routee-compass-macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "routee-compass-macros"
-version = "0.11.1"
+version = "0.11.2"
 edition = "2021"
 readme = "README.md"
 license = "BSD-3-Clause"
--- a/rust/routee-compass-macros/README.md
+++ b/rust/routee-compass-macros/README.md
@@ -10,7 +10,7 @@ To install as a library in Rust, add routee-compass-macros to your Cargo.toml fi
 
 ```toml
 [dependencies]
-routee-compass-macros = { version = "0.11.1" }
+routee-compass-macros = { version = "0.11.2" }
 ```
 
 This crate currently just provides a single macro for wrapping a compass application with python bindings.
--- a/rust/routee-compass-powertrain/Cargo.toml
+++ b/rust/routee-compass-powertrain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "routee-compass-powertrain"
-version = "0.11.1"
+version = "0.11.2"
 edition = "2021"
 readme = "README.md"
 license = "BSD-3-Clause"
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/routee-compass"
 exclude = ["test", "**/.DS_Store", "target/"]
 
 [dependencies]
-routee-compass-core = { path = "../routee-compass-core", version = "0.11.1" }
+routee-compass-core = { path = "../routee-compass-core", version = "0.11.2" }
 smartcore = { version = "0.4.0", features = ["serde"] }                       # random forest
 thiserror = { workspace = true }
 log = { workspace = true }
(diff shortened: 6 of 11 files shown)


python/nrel/routee/compass/io/generate_dataset.py
```diff
@@ -219,15 +210,15 @@ def replace_id(vertex_uuid: pd.Index) -> pd.Series[int]:
                 "nrel.routee.compass.resources", filename
             ) as init_toml_path:
                 with init_toml_path.open() as f:
-                    init_toml = toml.loads(f.read())
+                    init_toml: dict[str, Any] = tomlkit.load(f)
```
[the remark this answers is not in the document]
nreinicke: This was done to avoid mypy errors since the tomlkit library does not yet have robust mypy suppot. See [here](https://github.com/python-poetry/tomlkit/issues/326#issuecomment-1873372769).

robfitzgerald:
will return for a review next week, but wanted to reply to

> I was confused by the concept of an optional unit in the input feature and so I'm proposing that we update it to just use the baseunit if the units truly don't mater to the traversal model rather than making it optional

this is related to automated unit conversion, what we are building toward. as it is designed, some upstream traversal model is setting it's unit via an OutputFeature. that is the unit that the state model stores the feature value in when writing to the vector. when some downstream traversal model wants to retrieve the feature, it can specify the unit it wants the value in (the Some case) or it can accept the value in the unit it was stored in (the None case). 

currently, that downstream model will retrieve the feature using the `state_model.get_*()` and the unit value is provided _again_, so you're right, the InputFeature's optional unit isn't doing anything... _yet_. but we will need it, or something like it, in order to move toward automated unit conversion. i'll also mention, my point is true regardless of using uom or our homespun alternative.

robfitzgerald:
@nreinicke assuming i should wait on a review until changes are incorporated. should i also wait for a rollback on the optional input feature units?

rust/routee-compass-core/src/model/traversal/default/elevation/elevation_traversal_model.rs
```diff
@@ -40,7 +40,7 @@ impl TraversalModel for ElevationTraversalModel {
             ),
             (
                 String::from(fieldname::EDGE_GRADE),
-                InputFeature::Grade(Some(super::ELEVATION_GRADE_UNIT)),
+                InputFeature::Grade(None),
```
robfitzgerald, line 43: was this swapped for None by mistake? was `Some(super::ELEVATION_GRADE_UNIT)` which is Decimal

rust/routee-compass-powertrain/src/model/prediction/prediction_model_record.rs
```diff
@@ -21,130 +16,105 @@ pub struct PredictionModelRecord {
...
-                    None => {
-                        let (energy_rate, _energy_rate_unit) = self
-                            .prediction_model
-                            .predict((speed, speed_unit), (grade, grade_unit))?;
-                        cache.update(&key, energy_rate.as_f64())?;
-                        energy_rate
+        let (distance, distance_unit) =
+            state_model.get_distance(state, fieldname::EDGE_DISTANCE, Some(&self.distance_unit))?;
+        let mut feature_vector: Vec<f64> = Vec::new();
+        for (name, input_feature) in &self.input_features {
```
robfitzgerald, line 85: here we start to step through the declared set of input features set for this prediction model record. from your PR description, these changes
> [update] our vehicle models to accept N input features rather than hard coding them as speed and grade model

but right here, we restrict the possible types of input feature units to Speed and Grade. so, in a sense, this implementation really only supports those two units. i'm guessing to some degree, you're leaving the door open here to add Temperature and Mass in the future, and they should be updated here when that is the case. however, is there a reason to block `Time`, `Distance` and `Custom` `InputFeature` types here?

robfitzgerald (commented):
heya! excited to pave the way for mass and temp modeling. i have two questions in here that may be change requests depending on your reply. :guitar:

rust/routee-compass-powertrain/src/model/prediction/prediction_model_record.rs
nreinicke: I've added the ability to use more of the input features but right now if you add a model in that uses other features it will still block you when it tries to find [the minimum energy rate](https://github.com/NREL/routee-compass/blob/bbe6b7c2e3fbd0ce64a5be4468ed49c72852c55d/rust/routee-compass-powertrain/src/model/prediction/prediction_model_ops.rs#L54). There's not a clear path forward to generate sample values for something like Time, Distance and especially a custom variable and so maybe this opens the door to future work to allow the user to specify the sample values (or ranges) for each of the input features.

robfitzgerald (approved):
cool, thanks for the updates. looks good!

> removes the onnx feature; I would like to revisit this in the future but since we're not using it the code was a bit out of date with the latest ort runtime and would take some effort to build it back up.

that's a bummer but these things are moving quickly. hopefully the breaking changes from ort also signify runtime improvements for rust ONNX inference, that it might be worth a revisit someday, because only supporting smartcore is pretty limiting. thanks for cleaning that up.

PR #327, opened 2025-05-27
open constructors for TimeTraversalModel

robfitzgerald:
i discovered that the TimeTraveresalModel cannot be constructed outside of using a `TimeTraversalBuilder` build method, and this PR simply
  - gives it a new method to build from a TimeUnit
  - gives it a From method to build from a TimeConfiguration
  - makes TimeConfiguration public
  - makes the TimeTraversalModel::time_unit struct public


nreinicke (approved):
Good catch, this makes sense to me

PR #337, opened 2025-06-12
minor fixes to elevation traversal model

nreinicke:
When testing the elevation traversal model I was finding that the elevation gain and loss was always zero. This applies some minor fixes to get the trip elevation working:

 - update the grade input file key (it was just giving zero grades since the input path wasn't found)
 - update trip elevation loss to be distance base rather than grade based

I also made the grade model input required since it would prevent this silent error in the future. 

Lastly, this sets up our default energy configs to use the elevation traversal model

### DIFF
--- a/rust/routee-compass-core/src/model/traversal/default/grade/grade_configuration.rs
+++ b/rust/routee-compass-core/src/model/traversal/default/grade/grade_configuration.rs
@@ -5,10 +5,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct GradeConfiguration {
     /// file with dense mapping from edge id to grade value
-    /// note: this is optional as a stop-gap to allow for loading energy models that do not require grade.
-    /// this is because we currently cannot define input features that are _optional_, and so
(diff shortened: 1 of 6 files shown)


rust/routee-compass-core/src/model/traversal/default/grade/grade_configuration.rs
```diff
@@ -5,10 +5,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct GradeConfiguration {
     /// file with dense mapping from edge id to grade value
-    /// note: this is optional as a stop-gap to allow for loading energy models that do not require grade.
-    /// this is because we currently cannot define input features that are _optional_, and so
-    /// if a dummy grade model were not provided, energy models would fail to build.
-    pub grade_input_file: Option<String>,
+    pub grade_input_file: String,
```
robfitzgerald, line 8: > I also made the grade model input required since it would prevent this silent error in the future

this was a dumb stopgap, i think i agree it's not needed.

robfitzgerald (approved):
😸

PR #339, opened 2025-06-13
tag "type" for deserialization

robfitzgerald:
another small edit allowing a CustomFeatureFormat variant to be selected via a "type" tag, allowing:
```toml
format = { type = "floating_point", initial = 0.0 }
```


nreinicke (approved):
makes sense to me

PR #341, opened 2025-06-16
Update osm_default_energy_all_vehicles.toml

nreinicke:
A teeny tiny change to add elevation traversal model as default for osm default config with all vehicles.


robfitzgerald (approved):
This will really _elevate_ our routing algorithm.

PR #346, opened 2025-06-23
Benchmark CI

### LINKED ISSUE #336: Automated CI benchmark
In order to track the performance impact of changes to the repo, we can use GitHub actions designed for benchmarking and the rust criterion.rs library to
  - at minimum, record a history of performance as we make changes to main
  - explore defining a performance threshold to use as an acceptance criteria for code changes

brycemines:
Create a simple benchmark case suggested in #336 and create CI benchmarking workflow.

After the last discussion with @robfitzgerald, we might need to switch `secrets.GITHUB_TOKEN` to a different name.

closes #336

### DIFF
--- a/.github/workflows/rust-bench.yaml
+++ b/.github/workflows/rust-bench.yaml
@@ -1,0 +1,39 @@
+name: Rust Benchmark
+on:
+  push:
+    branches: [main]
+    paths-ignore:
+        - "docs/**"
+        - "**.md"
+
+permissions:
+    contents: write
+    deployments: write
+
+jobs:
+  benchmark:
+    name: Rust Performance Check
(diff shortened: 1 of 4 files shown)


nreinicke (approved):
This looks great, thanks for adding in! I think we should be okay with the github token since that's a token that gets set by the action runner when it's initialized (and not set at the repo level).

robfitzgerald:
@brycemines you can click "merge pull request". Let's visually confirm that the benchmark runs on merge and without failure, then we can go ahead and delete your branch.

robfitzgerald:
@brycemines looks like we need to still fix something. See [the action log](https://github.com/NREL/routee-compass/actions/runs/15878927061/job/44773690273) for details.

PR #348, opened 2025-06-25
fix failing GA

brycemines:
`output.txt` is created in the `./rust` working directory, so this PR specifies the correct output file path.


robfitzgerald (approved):
🤞

brycemines:
Maybe. The first error was the file didn't exist. Now it exists but just doesn't have anything.

PR #350, opened 2025-06-30
BEV charging

nreinicke:
Opening this to start thinking about adding charging logic into the routing algorithm for BEVs. 

To start, this is _very_ simple and just restricts any frontier where the vehicles battery falls below a defined threshold and also adds logic for a vehicle to charge to 100% any time it hits a node with a charging station. This is working at a very basic level (see the new 04_charging_station_example.py file for a demonstration) but, I think it could be improved before we include this as default behavior. 

Some ideas for improvement:
 - think about ways to make charging more intelligent (this logic means a vehicle always tries to charge) 
 - set behavior parameters like partial charging


rust/routee-compass-core/src/model/label/label_model.rs
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+    }
+}
+pub trait LabelModel: Send + Sync {
+    fn label_from_state(
+        &self,
+        vertex_id: VertexId,
+        state: &[StateVariable],
+        state_model: &StateModel,
```
robfitzgerald, line 43: Maybe we pass in the whole EdgeTraversal in place of VertexId and state here, which it contains along with the cost.

nreinicke: I thought about doing that but it felt awkward when setting up the initial label and the edge traversal only has the edge id rather than the vertex id (unless I'm overlooking something)

rust/routee-compass-core/src/model/label/label_model.rs
```diff
@@ -0,0 +1,47 @@
...
+    network::VertexId,
+    state::{StateModel, StateVariable},
+};
+
+#[derive(PartialEq, Eq, Hash, Debug, Clone, Serialize, Allocative)]
+pub enum Label {
+    Vertex(VertexId),
+    VertexWithIntState {
+        vertex_id: VertexId,
+        state: Vec<u64>,
```
nreinicke, line 17: Was just thinking maybe we should also have a variant if we know we're only using one int to cut down on having to allocate a vec for a single number 

```rust
VertexWithSingleIntState {
  vertex_id: VertexId
  state: usize
}
```

rust/routee-compass-core/src/model/label/label_model.rs
robfitzgerald: So, maybe what I mean (reading from my phone, probably missing some details here) is to pass in the `(VertexId, SearchTreeBranch)` tuple?

rust/routee-compass-core/src/model/label/label_model.rs
robfitzgerald: Sure, a few specialized instances like that could be smart since this is such a ubiquitous value type, it would certainly make a difference.

docs/examples/04_charging_stations_example.py
```diff
@@ -0,0 +1,84 @@
+# %%
```
robfitzgerald, line 1: for whatever reason, when i run `_convert_examples_to_notebooks.py` for this file, it just renders as one big cell, so i only saw the final plot at first. perhaps a simple formatting fix

python/nrel/routee/compass/io/charging_station_ops.py
```diff
@@ -0,0 +1,353 @@
+import importlib
```
robfitzgerald, line 1: there's a lot of code in here. how safe is it to rely on all of these string matches and regexes for deserializing the souce TSDC data, or do you know it to be robust? should we consider adding an issue to include more error checking on this?

robfitzgerald: should we add a feature guard here for people who didn't install "[all]" dependencies of compass?

python/nrel/routee/compass/resources/osm_default_charging.toml
```diff
@@ -0,0 +1,181 @@
...
+[cost.weights]
+trip_distance = 1
+trip_time = 1
+trip_energy = 1
+
+# Used for restricting travel after the battery is depleted.
+[frontier]
+type = "battery_frontier"
+
+[label]
```
robfitzgerald, line 38: i love it, nice to see how this can be configured.

rust/routee-compass-core/src/model/cost/cost_model_service.rs
```diff
@@ -50,7 +50,13 @@ impl CostModelService {
         // invariant: this hashmap dictates the list of keys for all subsequent CostModel hashmaps.
         let weights: Arc<HashMap<String, f64>> = query
             .get_config_serde_optional::<HashMap<String, f64>>(&"weights", &"cost_model")?
-            .map(Arc::new)
+            .map(|query_weights| {
```
robfitzgerald, line 53: good catch!

rust/routee-compass-core/src/model/label/label_model.rs
```diff
@@ -0,0 +1,47 @@
...
+use allocative::Allocative;
+use serde::Serialize;
+
+use crate::model::{
+    label::label_model_error::LabelModelError,
+    network::VertexId,
+    state::{StateModel, StateVariable},
+};
+
+#[derive(PartialEq, Eq, Hash, Debug, Clone, Serialize, Allocative)]
```
robfitzgerald, line 12: label should be moved into a separate label.rs file.

rust/routee-compass-core/src/model/label/label_model.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Label::Vertex(vertex_id) => write!(f, "Vertex({})", vertex_id),
+            Label::VertexWithIntState { vertex_id, state } => {
+                write!(f, "VertexWithIntState({}, {:?})", vertex_id, state)
+            }
+        }
+    }
+}
+pub trait LabelModel: Send + Sync {
```
robfitzgerald, line 40: doc comments here would be helpful to disambiguate a Label from a LabelModel + to instruct downstream implementations.

python/nrel/routee/compass/io/charging_station_ops.py
nreinicke: > should we add a feature guard here for people who didn't install "[all]" dependencies of compass?

Yep that's a good idea, I'll fix that. 

> there's a lot of code in here. how safe is it to rely on all of these string matches and regexes for deserializing the souce TSDC data, or do you know it to be robust? should we consider adding an issue to include more error checking on this?

Yeah these functions are just based off the sample of the data that I pulled but it would definitely be helpful to do a more robust analysis of the values and maybe add some unit tests to validate the extraction. I'll add in some unit tests and open an issue for doing a more comprehensive review of the possible options.

rust/routee-compass-core/src/model/label/label_model.rs
nreinicke: good call, I'll add those in

PR #353, opened 2025-07-03
Implemented option to return a compressed file when the provided filename ends in `.gz`

### LINKED ISSUE #170: ResponseSink::File should have gzip option
it should be possible to write outputs to either JSON or CSV files that are gzip'd, in order to greatly reduce space requirements for runs.

this could either be explicitly handled by adding a new field to ResponseSink::File, or, it could be implicitly handled by checking the output file location's suffix for a ".gz"

yamilbknsu:
In order to reduce the size of the output, it will be compressed using `flate2` if the filename ends in `.gz`. Since `GzEncoder<File>` implements `Write`, most of the code can still be used. The primary challenge of using `GzEncoder` instead of `File` is that the former needs a call to `finish()` when the writing is done. Therefore, my solution implements the `InternalWriter` enum which wraps `File` and `GzEncoder` and implements `finish`. `ResponseSink` now receives an `Arc<Mutex<InternalWriter>>` and calls `finish` in the `close` method.

One important caveat with this solution is that after calling `finish` on `InternalWriter::GzippedFile`, subsequent write attempts will panic. This is due to `GzEncoder<File>::try_finish()`, yet I couldn't find a way to use `GzEncoder<File>::finish()` because it requires ownership, which is an issue when using `Arc`.

I'm happy to discuss this further.

closes #170

### DIFF
--- a/rust/routee-compass-core/src/model/traversal/default/speed/speed_traversal_service.rs
+++ b/rust/routee-compass-core/src/model/traversal/default/speed/speed_traversal_service.rs
@@ -49,7 +49,8 @@ impl TraversalModelService for SpeedLookupService {
             }
             None => None,
         };
-        let speed_limit = speed_limit_tuple.map(|(speed_limit, max_speed_unit)| max_speed_unit.to_uom(speed_limit));
+        let speed_limit = speed_limit_tuple
+            .map(|(speed_limit, max_speed_unit)| max_speed_unit.to_uom(speed_limit));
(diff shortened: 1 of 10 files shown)


yamilbknsu:
While testing I realized there was a part of the code that was directly writing the header of the output as plain text. I moved that logic into the `InternalWriter` enum so that we could write the header to the file following the same compression encoding.

PR #365, opened 2025-07-22
fix: deploy-docs and rust-bench GA no longer conflict

brycemines:
This changes the `deploy-docs.yaml` and `rust-bench.yaml`. I also created the new branch `bench`, which is based on nothing and has the 4 benchmark results that we lost from the overwriting conflict.

* `deploy-docs`:  checks out the `bench` branch and moves its contents into `./docs/_build/html` before the GitHub Pages action.
* `rust-bench`: remove deploy privileges and write the output to `bench` branch.

closes #362

PR #367, opened 2025-07-25
hold smartcore at =0.4.0

robfitzgerald:
this PR holds the smartcore library dependency at exactly "=0.4.0". out-of-the-blue, dependency resolution for smartcore = "0.4.0" was found to install version 0.4.2 and that version has a breaking change on working with our serialized powertrain models, leading to [this CI failure](https://github.com/NREL/routee-compass/pull/365#issuecomment-3109955567) and failure to compile any version of compass not locked to 0.4.0.


nreinicke (approved):
dig it, thanks for fixing!

PR #369, opened 2025-08-01
fix release action to build sdist; bump to v0.12.1

nreinicke:
When we published v0.12.0, we dropped python 3.9 support but our github workflow was only building the source distribution if it saw python 3.9. So, for v0.12.0 there was no sdist released and therefore no conda release associated with that version. This fixes that and bumps the patch version.


robfitzgerald (approved):
Good catch!

PR #370, opened 2025-08-04
Fix diesel unit deserialization

nreinicke:
A tiny fix to correct how we were de-serializing the energy rate unit "gallons diesel/mile".

Also proposes adding the battery frontier model as default behavior.


robfitzgerald (approved):
Nice catch!

PR #371, opened 2025-08-11
test case for gal diesel/mile type

robfitzgerald:
created this test case while :brain: was forgetting [this happened](https://github.com/NREL/routee-compass/pull/370)


nreinicke (approved):
makes sense to me, thanks for adding!

PR #378, opened 2025-08-20
Rjf/inventory

robfitzgerald:
this PR integrates the [inventory](https://crates.io/crates/inventory) crate, which allows downstream libraries to inject builders at compile time. in order to leverage this, a downstream library will want to register builders in it's lib.rs file:

```rust
use routee_compass::app::compass::BuilderRegistration;
use inventory;
use std::rc::Rc;

inventory::submit! {
  BuilderRegistration(|builders| {
    builders.add_traversal_model("my_model".to_string(), Rc::new(MyModelTraversalBuilder {});
    // ... more here, call any add* methods on the builders object
    Ok(())
  })
}
```

use of this is demonstrated in the routee_compass::app::compass::compass_builder_inventory file.

along the way, a few breaking changes:
  - CompassAppBuilder renamed to `CompassBuilderInventory` to disambiguate from the builders that it contains and to clarify it doesn't follow the Builder design pattern 
  - CompassAppBuilder::default() method was removed in order to make a clearer distinction on how to create a `new` builder (with all injected values) or an `empty` one, if needed
    - it is still possible to build a CompassBuilderInventory entirely from it's add methods, for example for testing
  - CompassApp::try_from((&Config, &CompassAppBuilder)) replaced with CompassApp::new(&Config, &CompassAppBuilder)

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -9,6 +9,7 @@ members = [
 ]
 
 [workspace.dependencies]
+inventory = { version = "0.3.21" }
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = { version = "1.0.140", features = ["preserve_order"] }
 rayon = "1.10.0"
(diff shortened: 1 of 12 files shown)


nreinicke (approved):
This looks great, thanks for adding in!

PR #386, opened 2025-09-02
Rjf/extend graph

### LINKED ISSUE #184: working with transit modeling or different graph layers
in the case of transit modeling,
  - we need edges appended to the graph which connect station stops
  - we need a RoadClassFrontierModel to know the mapping from these edges to a "transit" class name
  - we need a traversal model that knows how to find the transit edges in the graph by id

appending to the graph could take a few forms:
  - the user needs to add the links to the edges-complete file before running
      - _cumbersome_
  - TraversalModelBuilder::build can append links to the graph
      - all traversal models needs to keep track of the offset index + number of edges associated with the traversal model
      - how do we inform the RoadClassFrontierModel what the edge ids are?
  - TraversalModelService has a register_edges method
      - similar to how we update the state model
      - in order to know what the index will become, we need to have the un-modified graph in scope to check it's `.size`

robfitzgerald:
This PR extends the Graph data structure in Compass so that it supports multiple and disjoint adjacencies over the same set of vertices. This is done by providing a vector of `EdgeList`s:

```rust
pub struct Graph {
    pub vertices: Box<[Vertex]>,
    pub edge_lists: Vec<EdgeList>,
}

pub struct EdgeList {
    pub adj: Box<[CompactOrderedHashMap<EdgeId, VertexId>]>,
    pub rev: Box<[CompactOrderedHashMap<EdgeId, VertexId>]>,
    pub edges: Box<[Edge]>,
}
```

Now any edge access requires an additional `EdgeListId` index value, such as Graph::get_edge():

```rust
pub fn get_edge(&self, edge_list_id: &EdgeListId, edge_id: &EdgeId) -> Result<&Edge, NetworkError>;
```

some notes on this change:
  - the added indirection and verbosity of declaring an EdgeListId allows us to keep our _O(1)_ lookup time complexity
  - FrontierModel, TraversalModel, and AccessModels may vary over EdgeLists (for instance, SearchInstance::traversal_models is a `Vec<Arc<dyn TraversalModel>>`, one for each EdgeList)
    - configuration will be impacted, an edge list id required in TOML for a given service type

remaining work:
  - [ ] SearchApp
  - [ ] CompassApp
  - [ ] builders, configurations with EdgeLists
  - [ ] rename Graph2/SearchInstance2, clean up originals
  - [ ] test

Closes #184.


yamilbknsu (approved):
This changes make sense to me. I think keeping (edge_list_id, edge_id) lookup is worth it and easier to work with in the future.

I also ran the tests locally and they succeed. Thank you!

Issue #343, opened 2025-06-16
Add gasoline and diesel as uom units


nreinicke:
Add in new Energy units for gallons/liters gasoline and gallons/liters diesel using the[ uom unit macro](https://docs.rs/uom/latest/uom/macro.unit.html). One these are implemented we can [update our to/from uom functions ](https://github.com/NREL/routee-compass/blob/0a401a5396449a4ad8d01392bae003d411544403/rust/routee-compass-core/src/model/unit/energy_unit.rs#L24) on the EnergyUnit.

PR #389, opened 2025-09-11
Add liquid fuel as uom units

yamilbknsu:
In this PR I'm addressing #343

We need to bump `uom` down to `0.36.0` because there is a [bug](https://github.com/iliekturtles/uom/issues/520) preventing the use of `unit!` in external crates.

### DIFF
--- a/rust/routee-compass-core/src/model/unit/mod.rs
+++ b/rust/routee-compass-core/src/model/unit/mod.rs
@@ -16,7 +16,7 @@ pub use as_f64::AsF64;
 pub use cost::{Cost, ReverseCost};
 pub use distance_unit::DistanceUnit;
 pub use energy_rate_unit::EnergyRateUnit;
-pub use energy_unit::{ EnergyUnit, gal_gas, gal_diesel, liter_gas, liter_diesel };
+pub use energy_unit::{gal_diesel, gal_gas, liter_diesel, liter_gas, EnergyUnit};
 pub use ratio_unit::RatioUnit;
 pub use speed_unit::SpeedUnit;
 pub use temperature_unit::TemperatureUnit;


robfitzgerald (commented):
thanks for starting this. i don't see any issue with stepping uom back a minor version. i have three questions about the implementation that aren't exactly requests for changes.

## extensibility - usage from downstream lib (bambam, etc)

say we want to interact with energy_unit from bambam. is it avaiable as we have written? do we continue using units from uom::si::f64, or, do we need to swap all references to a custom system exported from routee_compass? bonus points, can you show what that might look like in the downstream library?

## extensibility - adding more units

say we decide we want to add yet another custom unit type. someone defines/appends to the uom system in another file, for say some new type of TemperatureUnit. do the macro invocations in temperature_unit.rs and energy_unit.rs both update the global system? or would we need to create our new system in exactly one file in our code base, perhaps a routee_compass_core::model::unit::system module, so that the calls to unit! are recorded in a single, new system?

## GGE

can we add gallons gasoline equivalent and diesel gasoline equivalent to this list? conversion defined [here](https://www.nist.gov/system/files/documents/2016/09/27/08-st-apdx-e-14-annual-final.pdf). i think we can skip liters gas/diesel equiv. for now.

yamilbknsu:
Just adding a note, the following test passed (not sure if we want to put it in bambam or not):
```rust
#[cfg(test)]
mod tests{
    use routee_compass_core::model::unit::gal_gas;
    use uom::si::f64::Energy;
    
    #[test]
    fn tests_units(){
        let some_value = Energy::new::<gal_gas>(1.);
        print!("{:#?}", some_value);
        assert_eq!(some_value.get::<uom::si::energy::joule>(), 126832984.);
    }
}
```

robfitzgerald (approved):
Nice, thanks for answering my questions and confirming we can access the new units downstream. Have a good weekend!

robfitzgerald:
@yamilbknsu this PR is approved, but awaiting your merge

yamilbknsu:
Thanks for the heads up!

PR #397, opened 2025-10-03
Minor doc fixes

nreinicke:
Some minor doc fixes that I made while reviewing #396

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -16,6 +16,7 @@ parts:
       - file: examples/02_different_powertrains_example
       - file: examples/03_time_energy_tradeoff_example
       - file: examples/04_charging_stations_example
+      - file: examples/05_ambient_temperature_example
 
   - caption: User Reference
     chapters:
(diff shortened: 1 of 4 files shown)


nreinicke:
Sorry for not getting back to you - was having some technical difficulties. 

I just went back and reviewed our `generate_dataset` function and it looks like the change I made would actually cause the code to break in the case where the GRADE pipeline step was not included since it won't download the grade and [save it to a file](https://github.com/NREL/routee-compass/blob/986187b19b6ba330ad562b370c46bc2e127ca289/python/nrel/routee/compass/io/generate_dataset.py#L201). Given that grade is very important for any of the energy models, how would you feel if I just removed the GRADE pipeline step entirely and just made it a required step to do the grade download?

robfitzgerald:
> Given that grade is very important for any of the energy models, how would you feel if I just removed the GRADE pipeline step entirely and just made it a required step to do the grade download?

that's a good stop gap measure. or, do we just run the GRADE logic if the POWERTRAIN phase is active, and remove the GRADE phase?

nreinicke:
> that's a good stop gap measure. or, do we just run the GRADE logic if the POWERTRAIN phase is active, and remove the GRADE phase?

Thanks a good idea, I've just made a change to only do the grade download if we have the powertrain phase active and have removed the grade step

PR #401, opened 2025-10-16
make all CompassAppConfig types accessible from downstream crates

robfitzgerald:
this small PR exposes EdgeListConfig and SearchConfig at the module so it can be used by downstream libraries when working with CompassAppConfig instances.


nreinicke (approved):
makes sense to me

PR #404, opened 2025-10-23
Rjf/403 graph adjacencies

### LINKED ISSUE #403: EdgeLists containing all adjacencies does not scale
we are running bambam with 500 edge lists and running into a persistent OOM error on Kestrel.

this is caused by the fact that each edge list [creates its own pair of dense adjacency lists](https://github.com/NREL/routee-compass/blob/main/rust/routee-compass-core/src/model/network/edge_list.rs#L27).

to address this, the adjacencies should be moved up a level to the Graph and should be keyed by EdgeListId and EdgeId.

robfitzgerald:
Closes #403 by pulling the fwd/rev adjacency lists out of each edge list and instead holding one forward + reverse adjacency list across the whole graph. changes the keys of the adjacency lists from EdgeId to (EdgeListId, EdgeId). this removes the unnecessary duplication of dense adjacency list vectors which causes memory bloat when many edge lists exist. it should also improve runtimes for calling `Graph.{in|out}_edges_iter` as we no longer need to scan all edge lists. no compass APIs have been affected.

### DIFF
--- a/rust/routee-compass/src/app/cli/run.rs
+++ b/rust/routee-compass/src/app/cli/run.rs
@@ -5,7 +5,6 @@ use crate::app::compass::{
 };
 use itertools::{Either, Itertools};
 use log::{debug, error};
-use routee_compass_core::config::CompassConfigurationError;
 use serde_json::{json, Value};
 use std::io::BufRead;
 use std::{fs::File, io::BufReader, path::Path};


robfitzgerald:
> Right, this is way more efficient to store it like this, good catch.

Oh it is! With 480 transit graph edge lists I was getting OOM errors on Kestrel with a 950G RAM node. Now it runs normally.

PR #405, opened 2025-10-27
support python 3.14

nreinicke:
Some minor updates to support python 3.14. The bulk of the changes come from adding support for the pixi tool to manage multiple python environments locally for testing.


nreinicke:
I may have jumped the gun on this one. Looks like rasterio has not yet published wheels for py314 via PyPI. I can successfully install via pixi which uses the conda-forge packages but if we still want to support a basic `pip install` we might want to hold off until we see a py314 wheel from rasterio. I think I'll close this for the time being and open a new PR with just the pixi addition.

PR #407, opened 2025-10-28
Rjf/graph progress

robfitzgerald:
some small changes to progress bars during graph dataset imports.


nreinicke (approved):
This makes sense to me

PR #413, opened 2025-10-30
allow newline json without chunksize

### LINKED ISSUE #412: CLI should support newline delimited without chunksize
if `--newline-delimited` is provided when running compass but no `--chunksize` is specified, the program fails. however, if parallelism is 1 and the inputs are newline-delimited, it may not make sense to define a chunksize.

error message:

```
[2025-10-30T16:35:37Z ERROR mep] internal error: invalid argument combination should have been caught during CLI validation
```

robfitzgerald:
Closes #412 by allowing the `--newline-delimited` flag without specifying a `--chunksize` argument. if no chunksize is specified, chunsize is set to MAX (one big chunk).

### DIFF
--- a/rust/routee-compass/src/app/cli/run.rs
+++ b/rust/routee-compass/src/app/cli/run.rs
@@ -50,17 +50,14 @@ pub fn command_line_runner(
 
     // execute queries on app
     match (args.chunksize, args.newline_delimited) {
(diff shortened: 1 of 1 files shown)


nreinicke (approved):
makes sense to me

PR #415, opened 2025-11-04
Rjf/termination without failure

### LINKED ISSUE #414: TerminationModel allowing termination without failure
When routing from an origin to build a search tree, it may be desirable to allow the planner to return the tree, even when the search does not result in a natural termination within the termination model window. For example, if the traversal model + cost model is unstable, it may result in a local minima and the frontier queue may oscillate. This may still produce an acceptable tree result.

To allow this, each `TerminationModel` variant could have an optional `TerminationBehavior` argument, something like:

```rust
pub enum TerminationBehavior {
  FailureOnTermination,
  SuccessOnTermination,
}
```

### notes

Does the `TerminationBehavior` live as a member of each variant, or, should it be a setting of the SearchAlgorithm instead?

When routing a point-to-point request, aka, when a destination is provided, it is never sensible to allow for a termination model to "return anyway". This fact should be codified in the logic and trigger an application build error instead of a search error.

robfitzgerald:
this PR closes #414 by introducing a TerminationFailurePolicy configurable for instances of the A* SearchAlgorithm. two variants are supported: AllTerminationsFail and PathTerminationsFail. depending on the value set and the result of the search, the program will either propagate up the termination explanation as an error (AllTerminationsFail case) or, in the case that the search did not have a destination, allow the result to be returned (AllowTreeTermination case). the policy is an optional argument for Dijkstra and A* SearchAlgorithm variants; if not included, the default (AllowTreeTermination, most permissive) is used.

along the way,
  - added a memory limit termination function
  - _modernized_ the termination module


nreinicke (approved):
This looks good. I've tested this locally and it exhibits the expected behavior. How would you feel about also updating the [example config in the docs ](https://nrel.github.io/routee-compass/config.html) to add in a section showing an example of setting the termination condition and behavior?

PR #422, opened 2025-11-18
Rjf/dependencies

robfitzgerald:
this PR improves dependency management and updates all dependencies where possible for compass. this includes:
  - moving all versioned dependency references into the workspace TOML (all dependencies in crates are "workspace = true"
  - sorting dependencies by name
  - CI testing for sorted dependencies on PR
    - devs can use `cargo sort` via `cargo install cargo-sort` which automates this
  - updating all dependencies to latest (except uom and smartcore where we have known issues)
  - use of [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates) 

edit: had to add some special handling around ninterp 0.8 and ndarray to avoid multiple ndarray versions in the cargo tree

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,13 @@ updates:
       directory: "/rust/"
       schedule:
         interval: "weekly"
+      cooldown:
+        default-days: 7
 
     # GitHub Actions versions
     - package-ecosystem: "github-actions"
       directory: "/"
       schedule:
         interval: "weekly"
+      cooldown:
+        default-days: 7


nreinicke:
This makes sense to me. I want to hear @kylecarow's take on the ninterp ndarray fix since I think he has been dealing with this on fastsim too.

robfitzgerald:
> This makes sense to me. I want to hear @kylecarow's take on the ninterp ndarray fix since I think he has been dealing with this on fastsim too.

Right, I would too. @kylecarow I got around it for us by using the notation "=0.17.1" in our cargo workspace file for ndarray. Are you doing something similar? Is it happening because ndarrays are in the function signatures/API for ninterp?

robfitzgerald:
adding a dependency cooldown after reading [this article](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns)

Issue #417, opened 2025-11-04
update RAM reporting using custom Allocative implementations


robfitzgerald:
after switching to the SearchTree, we have seen that the `allocative::size_of_unique` method does not work correctly for deeply-nested types that simply derive Allocative. calling `allocative::size_of_unique` on _any_ SearchTree will yield a size of 144 bytes _every time_.

to address this, we need to manually implement Allocative for SearchTree and for SearchAppResult. this follows the visitor pattern, where we must visit each object that we wish to report. along the way, we need to figure out which inner objects properly report their size and which do not in order to avoid double-counting.

PR #437, opened 2025-12-16
Rjf/optional allocative

robfitzgerald:
this PR allows users of the summary output plugin to include memory estimation (off by default) by setting estimate_memory_consumption = true:

```toml
[[plugin.output_plugins]]
type = "summary"
estimate_memory_consumption = true
```

does not address standing issue with Allocative and SearchAppResult described in #417.


nreinicke (approved):
This makes sense to me; good to have it as optional

PR #449, opened 2025-12-24
Config filepath fixes

nreinicke:
A small fix to how we normalize paths to allow any output files to pass through the normalization process without causing an error.

Also fixes a small config error in the osm_default_charging.toml file

### DIFF
--- a/rust/routee-compass-core/src/config/config_json_extension.rs
+++ b/rust/routee-compass-core/src/config/config_json_extension.rs
@@ -296,16 +296,13 @@ impl ConfigJsonExtensions for serde_json::Value {
     }
 
     /// This function is used to normalize file paths in the configuration JSON object.
-    /// Incoming file paths can be in one of three locations:
+    /// Incoming file paths can be in one of two locations:
     ///
     /// 1. Absolute path
-    /// 2. Relative path to the config file itself
-    /// 3. Relative path to where the application is being run
+    /// 2. Relative path to the config file
     ///
     /// This function scans each key value pair in the config and for any key that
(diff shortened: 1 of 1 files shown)


rust/routee-compass-core/src/config/config_json_extension.rs
```diff
@@ -386,12 +384,10 @@ impl ConfigJsonExtensions for serde_json::Value {
...
-                        // if we can't find the file in either location, we throw an error
-                        Err(CompassConfigurationError::FileNormalizationNotFound(
-                            String::from(parent_key.as_ref()),
-                            path_string.clone(),
-                            new_path_string,
-                        ))
+                        // if the file doesn't exist in either of these locations, it's possible that it could speficy an output file that doesn't exist yet
+                        // if this is the case, we just return the path as-is.
+                        // For input files that aren't found, an error will be raised later when attempting to read the file.
+                        Ok(serde_json::Value::String(path_string.clone()))
```
robfitzgerald: but if we do this, then we don't get immediate feedback when reading a config that a file doesn't exist. if there's 10 minutes of traversal models, constraint models, input plugins, and then an erroneous input file as part of an output plugin (i'm looking at _you_, [bambam](github.com/NREL/bambam)), then we lose that feedback, which is a good deal of why we do this normalization and file check. 

that said, this isn't the first time i've visited this code in the last few months. i'm wondering if we could just generally do better. what do the LLMs think of this? do they recommend a best practice and/or library for this?

robfitzgerald (approved):
i like the added feature of array input file(s) validation. :rocket:

PR #452, opened 2026-01-06
Performance updates

nreinicke:
I was struggling with memory errors when running a large scale performance analysis and so I did a detailed performance and memory audit on compass and picked off several lowish hanging fruits for improving the memory efficiency and the runtime performance. These updates significantly reduce the memory consumption of the search tree. For a sample regional query that has a trip distance about 1800 miles, the search result mib size was reduced from ~18000 to ~1100. For this same query, the runtime was reduced from 5:10 minutes to 2:14 minutes. Here is a brief summary of the updates:

## summary of updates

 - Don't keep track of the cost components for each edge traversal during the search (24b75eeb51bd8639277d196da83a55b5fd98927c). These cost component dictionaries were using string keys for all of the component costs on an edge and were taking up a lot of memory. These components can be deduced from the edge states after the fact and so I think it would be much more efficient to add these back in as an optional output plugin rather than storing them during the search.
 - Removes some components and sub components of the search result that were not being used (56e984851f781323275bb7c27c1bb557456a6d23).
 - Allow trip level state variables to be omitted (220faee9378d3dca3bf886bbe3bcd081f4e44c19). Since trip level metrics can be inferred from edge metrics after the fact, I made them optional at the config level. This reduces the size of the search state vector.
 - Update the label enum to store the state on the heap (239e77cffbd4a53f40be9f0c6edb974988b4710d). This allows the simpler variants (like vertex id) to be about half the size.
 - Don't keep track of mapping from Label -> VertexId for the Label::VertexId variant since it's redundant (65c02b821fcba0ffcde3d8e59133b8698a1cd691)
 -  Keep track of if a cost feature is an accumulator or not rather than looking it up in a hash map at every edge traversal (519f564113296f14ad56b53d5ad797881346c44d)
 - Encode road classes as a u8 from the original string representation to reduce memory and hash lookup time (337423de78597b49ef5a42bce2b72c6171194f0d)

### DIFF
--- a/rust/routee-compass-core/src/algorithm/search/search_tree.rs
+++ b/rust/routee-compass-core/src/algorithm/search/search_tree.rs
@@ -49,6 +49,8 @@ impl SearchTree {
     pub fn set_root(&mut self, root_label: Label) {
         let root_node = SearchTreeNode::new_root(self.direction);
         self.nodes.insert(root_label.clone(), root_node);
+        // Note that we only introduce non-vertex labels into the labels map since,
+        // the Label::Vertex variant is effectively equivalent to the vertex ID itself.
         if !matches!(root_label, Label::Vertex(_)) {
             self.labels
                 .entry(*root_label.vertex_id())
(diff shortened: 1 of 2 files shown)


nreinicke:
@robfitzgerald - As a heads up, I did add the cost component tracking as an optional feature for your consideration when you get a chance to review. If we do go this route I might leave the small string representation as future work

PR #455, opened 2026-01-12
Ndr/performance

nreinicke:
After a bit more review, I've cherry picked some of the performance upgrades in #454 here to include some that have had measurable gains when testing in a large scale batch run. Those are:

 - Use specialized ninterp 1/2/3d classes when we have a smaller feature space.
 - Get previous edge id without fill backtrack
 - Add parquet writer to write to parallel parquet partitions (this one actually didn't measurably improve the overall runtimes but it didn't hurt them either so kept this in as a nice option to have)


robfitzgerald (approved):
looks great! thanks for all of your wrangling! :beers:

PR #464, opened 2026-02-04
Add state to map matching results

nreinicke:
Extends upon the map matching results to also include the search state in the matched path and a final traversal summary in the map matched result. Now, when map matching, a user can quickly get trip or edge level metrics like total energy consumed or total travel time. 

Note that this update changes the serialization of the traversal summary a bit. Now, we encode which summary op was used and what unit the variable has. For example:

```json
{
    "edge_speed": {
        "value": 42.02192799852017,
        "unit": "miles/hour",
        "op": "avg"
    },
    "edge_energy_liquid": {
        "value": 0.40017462491453454,
        "unit": "gallons_gasoline_equivalent",
        "op": "sum"
    },
    "trip_elevation_gain": {
        "value": 0.0328126010941803,
        "unit": "miles",
        "op": "last"
    }
}
```

This update also required a new method on the search instance `recalculate_path` that can take an existing path (like the one we get back from map matching as a vector of edge ids) and recompute the state transitions. This is needed since the LCSS algorithm compute paths recursively and then joins them back without state context. This ensures that accumulators like `trip_time` are correct in the final result.

Lastly, I made some fixes the examples and python utilities to be compatible with these changes.

### DIFF
--- a/rust/routee-compass/src/app/compass/compass_app.rs
+++ b/rust/routee-compass/src/app/compass/compass_app.rs
@@ -9,24 +9,20 @@ use crate::{
     app::{compass::CompassAppError, search::SearchApp},
     plugin::{input::InputPlugin, output::OutputPlugin},
 };
-use chrono::Local;
 
 use kdam::Bar;
 use rayon::current_num_threads;
-use rayon::prelude::*;
 use routee_compass_core::algorithm::search::SearchAlgorithm;
 use routee_compass_core::model::cost::cost_model_service::CostModelService;
 use routee_compass_core::model::map::MapModel;
-use routee_compass_core::model::network::{EdgeId, EdgeListId, Graph};
+use routee_compass_core::model::network::Graph;
 use routee_compass_core::model::state::StateModel;
-use routee_compass_core::util::duration_extension::DurationExtension;
 use serde_json::Value;
 use std::{
     path::Path,
     sync::{Arc, Mutex},
 };
 
-use crate::app::map_matching::{MapMatchingAppError, MapMatchingRequest};
 use routee_compass_core::algorithm::map_matching::MapMatchingAlgorithm;
 
 /// Instance of RouteE Compass as an application.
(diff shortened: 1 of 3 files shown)


robfitzgerald:
@nreinicke noting your request for review but also un-addressed reviews from our robot overlords, did you want me to continue to wait for those changes?

rust/routee-compass-core/src/model/cost/traversal_cost.rs
```diff
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use crate::model::unit::Cost;
 
 /// the cost of an edge traversal.
-#[derive(Serialize, Deserialize, Default, Clone, Debug, Allocative)]
+#[derive(Serialize, Deserialize, Default, Clone, Debug, Allocative, PartialEq)]
```
robfitzgerald: just checking here, a derived PartialEq will check for equality on all fields of TraversalCost. was this the intent, or, do we only need to compare objective_costs between TraversalCost instances?

rust/routee-compass-core/src/model/state/state_variable_config.rs
```diff
@@ -112,6 +112,31 @@ impl StateVariableConfig {
         }
     }
 
+    pub fn get_unit_name(&self) -> String {
```
robfitzgerald: this seems to duplicate the functionality above. could we 1) delete this, 2) give it doc comments and explain why we need it? if we do need it, could we give these two functions clearer names and also share implementation where possible?

rust/routee-compass/src/app/search/route_output.rs
```diff
@@ -0,0 +1,163 @@
+use crate::plugin::output::default::traversal::TraversalOutputFormat;
```
robfitzgerald: let's break up this file into parts, perhaps into a module named route/

rust/routee-compass-core/src/model/state/state_variable_config.rs
nreinicke: Yes, you're right, I will consolidate.

PR #467, opened 2026-02-11
Update organization names in README and License

nreinicke:


robfitzgerald (changes requested):
Good catch! One comment from my unreliable memory of our original TTO discussions. Also, could you update the LICENSE.md file as well?

PR #473, opened 2026-02-19
Rjf/revise config type handling

robfitzgerald:
this PR refactors the handling around "type" keys in traversal and constraint configurations. instead of the requirement leaking through to each Config struct as an expected key, it is instead stripped from the config before it gets passed through.


nreinicke (approved):
Yep, this is nicer, cleaner way to handle this, thanks for updating!

PR #480, opened 2026-03-03
address error cases where transparent error propagation creates ambiguity

robfitzgerald:
this tiny PR adds error message prefixes to some errors which previously used transparent propagation, in an effort to provide additional context on these errors. the transparent keyword is still used when no additional context is required.


nreinicke (approved):
Looks good, thanks for updating!

PR #488, opened 2026-03-31
SearchTree renamed SearchGraph

robfitzgerald:
ever since we introduced the LabelModel, search is no longer guaranteed to produce a tree. this PR renames SearchTree to SearchGraph. the description describes it as a directed acyclic graph (DAG). this covers the wider definition of what we build in cases where a node may have multiple parents.

this is mostly to align naming conventions with the upcoming journal paper.


robfitzgerald:
@nreinicke regarding multiedge vs optimal edge: the default behavior is to plot _all edges_ of a search graph. i'm just going to leave it there for this PR and if we want to, we can revisit `traversal_ops::create_graph_multilinestring` and pick the top-scoring Label for each VertexId when grabbing all edges.

robfitzgerald:
this led to the IRL discussion and assertion that, in fact, _no_, in fact in fact, Compass generates a tree after all.

PR #492, opened 2026-04-02
fixing incorrect pruning + backtrack handling in SearchTree

### LINKED ISSUE #491: search pruning should order labels, instead should prune nodes grouped by labels
in the current implementation of search_pruning.rs, we maximize over the label space and minimize over the cost space in order to determine which labels can be safely pruned. while makes sense when a label type has an ordering, it is incorrect both in this case and the categorical case, as the objective in label-based search is to allow disjoint label spaces to effectively behave as disjoint trees, not as a directed acyclic graph (DAG). for our search to be correct, we should instead treat labels at the same vertex but with different label state to be effectively separate search tree nodes.

robfitzgerald:
this PR set out to address #491, where nodes are being removed by a combination of maximizing state and minimizing cost. this has the effect of incorrectly managing subtree spaces, as the tree spaces associated with two different label states at the same vertex should not be treated as related in the search space. however, the impact of investigating this change led to the discovery of additional logic fixes.

**1. Removed Tree Pruning for Simplified Insertion Logic**
The complex `search_pruning.rs` module and Pareto-dominance pruning logic were removed entirely. `SearchTree` assumes testing for cost improvement occurred in the search algorithm. This significantly simplifies tree insertion while actively preventing cycles.

**2. Introduced `EdgeTraversalContext` for Traversal Models**
We needed to add the parent Label of a traversal in order to know where to start any backtracking. This is a requirement of the TurnDelay TraversalModel. It was incorrectly grabbing the min cost label associated with the source Vertex of a traversal. by passing in the parent Label, we can index into the correct subtree.

This increased the number of arguments to traverse_edge. To clean up `TraversalModel::traverse_edge` and `CostModel` method signatures, the loose trajectory arguments `(&Vertex, &Edge, &Vertex)` and `&SearchTree` along with the parent `Label` were bundled into a comprehensive `EdgeTraversalContext`. This context also provides safe helper methods like `backtrack()` and `previous_edge_traversal()`, reducing boilerplate in models that look into the search history (like `TurnDelayTraversalModel`). 

**3. Refactored `TraversalCost` and Enforced Algorithmic Cost Constraints**
`TraversalCost` was revised to be more semantic: `total_cost` was renamed to `edge_cost` to clarify that it defines the marginal cost of an edge—not the accumulated path cost. Additionally, a `CostConstraint` enum (`StrictlyPositive` vs `Unconstrained`) was wired from the top-level `SearchAlgorithm` configuration down through the `CostModel` and into `TraversalCost::insert()` to ensure algorithms like A*/Dijkstra safely enforce monotonic (positive) costs, leaving an opening for any algorithms in the Bellman-Ford family which do not require monotonicity.

---

### Minor Changes

**`SearchTree` & Algorithms:**
* Added a recursive `calculate_total_objective_cost` to `SearchTree` that includes infinite-loop guards against malformed tree cycles.
* Renamed `SearchTree::insert()` to `SearchTree::insert_trajectory()`. 
* Added `get_min_accumulated_cost_label()` finds the `Label` at some `VertexId` with the least cost. 
* Removed the `child_count` property and associated methods from `SearchTreeNode`.

**`TraversalCost`:**
* Renamed `total_cost` field to `edge_cost` system-wide.
* Replaced `TraversalCost::default()` and `zero()` with clearer explicit constructors like `TraversalCost::empty()` and a `#[cfg(test)] pub fn mock()` helper.
* Integrated the new `CostConstraint` parameter into `TraversalCost::insert()`.
* Created `AccumulatedTraversalCost` type, constructed over a route and serialized in the traversal output plugin. This separates the edge-level costs at search time (`TraversalCost`) from the trip-level costs at output time (`AccumulatedTraversalCost`)

**`TraversalModel` & `CostModel` implementations:**
* Updated all core traversal models (`DistanceTraversalModel`, `SpeedTraversalModel`, `TurnDelayTraversalModel`, `SimpleChargingModel`, etc.) to use `EdgeTraversalContext` for edge and vertex data access.
* Adjusted `NetworkCostRate::network_cost` to pull trajectory details directly from `EdgeTraversalContext`.

Closes #491.


rust/routee-compass-core/src/algorithm/search/a_star/frontier_instance.rs
nreinicke, line 1: Now that we've simplified this, clippy caught that we can remove the outer loop in the pop_new method. We don't need to do this here but eventually we should add clippy checks into our rust actions.

rust/routee-compass-core/src/algorithm/search/search_tree.rs
nreinicke, line 1: I'm wondering if we should also remove the `get_min_cost_label` and `min_cost_ordering` functions since they return the label with the minimum _edge_ cost and not accumulated cost which could be misleading. Or, if you need this for downstream work, perhaps we do need to add cumulative cost to our search tree node so we can get the proper label out of this function.

nreinicke (approved):
This looks good and I love that you've simplified this. I have a couple of comments but will approve so you can proceed with downstream usage.

rust/routee-compass-core/src/algorithm/search/search_tree.rs
robfitzgerald: this is tricky, this is also core to how we backtrack, as that starts from a VertexId, but we need to select the Label associated with that destination VertexId that is least cost, and then backtrack through that tree. we in fact _do need this to be_ the accumulated cost for this selection to be correct. i'll get to work proposing a fix so that TraversalCost holds both edge_cost and total_cost (or trip_cost).

rust/routee-compass-core/src/algorithm/search/a_star/frontier_instance.rs
robfitzgerald: i think my rust toolchain just needed to be upgraded, i usually run clippy every time. but ya, i thought we were checking clippy in our CI workflow!

robfitzgerald:
@nreinicke this is passing now, yay, but, _it's a way different set of changes since your review_. i updated the description above, i would start there. i will directly reference this branch for my development tasks next week so that you can take your time to review these major code changes. have a great conf!

nreinicke (commented):
Overall this is looking good, I like that you're getting the accumulated cost at the end rather than storing it in the tree.

I was thinking a little bit more about the label storage in our SearchTree.label data structure and I think that we should retain the previous logic to not insert the redundant label if we're using the Label:Vertex variant. Based on napkin math, if we have 1M vertices in our search tree and we're storing the `labels` with Label:Vertex completely redundantly, we might be inflating our memory from 138 MB to 612MB (5x) for the search tree. This seems to me like a small amount of additional complexity (just don't store Label:Vertex variants in the labels map) more than makes up for the gain in memory performance.

robfitzgerald:
> I was thinking a little bit more about the label storage in our SearchTree.label data structure and I think that we should retain the previous logic to not insert the redundant label if we're using the Label:Vertex variant.

@nreinicke i went ahead and did it. and i wrote up a really nice summary comment here to describe it, but it got gobbled up when i changed page for a second. 😮‍💨 

so, here's an AI summary:

* **Introduced `TreeStorage` Abstraction**: Created a new internal enum bounding the data structures for `SearchTree`, separating the highly optimized `VertexOnly` variant (using a single `HashMap`) from the flexible `Stateful` variant (using multiple maps to handle complex paths).
* **Automatic Storage Evaluation**: Refactored `SearchTree::with_root` and `set_root` to inspect the initial `Label`. The tree now automatically morphs into the optimal memory structure based on whether a simple `Label::Vertex` or a stateful label is detected.
* **Strict Homogeneous Enforcement**: Implemented the new `SearchTreeError::HeterogeneousLabelTypes` error to ensure the tree strictly rejects any attempts to insert a mixture of simple vertex labels and stateful labels, enforcing the expectation that a `LabelModel` remains consistent. 
* **Unified Iterator API**: Designed the inner iterators for `VertexOnly` storage to automatically "lift" internal `VertexId`s back into `Label::Vertex` instances on the fly. This prevents calling code from needing to know which storage variant is currently active.
* **Propagated Result Handling**: Updated consumer code, including `a_star_algorithm.rs` and internal unit tests, to properly utilize the `?` operator for gracefully handling the new `Result` returns from tree initialization and trajectory insertions.
* **Comprehensive Unit Testing**: Added dedicated tests to the `tree_storage.rs` module that validate initialization, successful homogeneous insertion for both storage variants, and the expected rejection of heterogeneous label injections.

i made these changes _here_, which i thought was a good idea at the time, since your comment seems to imply that we should make this fix immediately. i hope that doesn't complicate any follow-up review. you can focus on SearchTree/TreeStorage for most of the details.

robfitzgerald:
i'll leave this for now, i guess it's not passing tests, at least on my local, a test of calculate_path logs this error:

```
thread 'app::compass::compass_app::tests::test_run_calculate_path' (11423591) panicked at routee-compass/src/app/compass/compass_app.rs:450:13:
String("internal error due to search logic: LabelModel must produce a consistent Label type, but produced VertexWithIntState { vertex_id: VertexId(1), state: 0 } after initially producing Label::Vertex")
```

this looks like an error in the app code itself, so i'll review (monday). also a host of map matching tests fail.

nreinicke (approved):
Took a look through this again and I think this looks good. Thanks for adding the tree storage enum, it's a clean way to manage the special optimized vertex only label condition.

PR #493, opened 2026-04-07
Rjf/state model contains key

robfitzgerald:
super little PR here to adjust the signature of `state_model.contains_key` so that it takes a &str instead of a &String. to get around allocating the string "route_id" at every search iteration in bambam, i need to use a [static allocation of a String via OnceLock](https://github.com/NatLabRockies/bambam/pull/126/changes#diff-38b7248da455d2f233f5ef79a66403e803646dd6eaa6afa982e3b268ac9347afR2), which is silly.


nreinicke (approved):
Yep, makes sense to me

PR #501, opened 2026-04-15
Rjf/constraint model context

### LINKED ISSUE #381: FrontierModel should have scope of entire search tree
in #350, the API for FrontierModel [was modified to remove the tree](https://github.com/NREL/routee-compass/commit/3474b9140565b92c87ca584d06526bb30436f864#diff-09407a8ceb16ae0248433770c01b670cea76ae5f4ab67e1f3b5d04f1e038b58bR29). it was reverted to only include the scope of the current and previous link. however, certain desirable frontier modeling features like maneuver-based filtering are not possible without the full tree history. 

In order to support maneuver filtering and other frontier restrictions that require more access to route history, we need to reinstate the tree as part of the frontier model API, in particular, reintroducing **tree** and **direction**:

```rust
    fn valid_frontier(
        &self,
        edge: &Edge,
        state: &[StateVariable],
        tree: &HashMap<VertexId, SearchTreeBranch>,
        direction: &Direction,
        state_model: &StateModel,
    ) -> Result<bool, FrontierModelError>;
```

robfitzgerald:
this PR refactors the ConstraintModel to accept the EdgeFrontierContext [^1] in place of the current and previous edge. this gives constraint algorithms the same scope as the TraversalModel:

```rust
    fn valid_frontier(
        &self,
        ctx: &EdgeFrontierContext,
        state: &[StateVariable],
        state_model: &StateModel,
    ) -> Result<bool, ConstraintModelError>;
```

along the way, 
  - added error variants for ConstraintModelError wrapping SearchTreeErrors and StateModelErrors for convenient error propagation from the usual suspects.
  - refactored `EdgeTraversal::new_local` and added new constructor for `EdgeFrontierContext` for better ergonomics
  - same for TraversalModelError with SearchTreeError.
  - renamed modules containing Traversal types.

Closes #381.

[^1]: formerly known as `EdgeTraversalContext`, renamed to avoid confusion with the existing `EdgeTraversal` type associated with the `TraversalModel`


nreinicke (approved):
This makes sense to me and is reads a lot cleaner than the previous disparate apis. Thanks!

PR #507, opened 2026-04-28
make powertrain dependency optional in routee-compass

robfitzgerald:
this PR makes it possible for downstream users like bambam to import routee-compass without powertrain/onnx compilation via `default-features = false`. in the builder inventory file, this skips importing routee-compass-powertrain and skips injecting the builders related to powertrain modeling.


nreinicke (approved):
This makes sense and I appreciate that it's a small lift to make it happen.

PR #516, opened 2026-05-26
added OneOrMany methods for construction + Vec-like push method

robfitzgerald:
this PR adds a few more methods for working with OneOrMany. 

### `From<Vec<T>>`

this new constructor will determine whether to unpack the Vec in the special case that the Vec has a length of 1 (`OneOrMany::One` case)

### empty, one, and many constructors

these constructors can be a more expressive way to work with the different underlying representations of OneOrMany.

### push

this method is like Vec::push in that it adds a single item to the end of the collection. here, it handles the special transitions between empty, one, and many values without cloning.

### tests

added 3 more tests to check promotion between OneOrMany types when calling the push method.

### DIFF
--- a/rust/routee-compass-core/src/config/one_or_many.rs
+++ b/rust/routee-compass-core/src/config/one_or_many.rs
@@ -102,7 +102,7 @@ impl<T: Clone> OneOrMany<T> {
         OneOrMany::One(t)
     }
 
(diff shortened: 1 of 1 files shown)


robfitzgerald:
@nreinicke this could use a review if you have the time. :wave:

nreinicke (approved):
So sorry for my delay in getting to this, it looks good

PR #531, opened 2026-06-17
New docs page for MacOS installation troubleshooting (pyo3 linking errors)

admrtin:
This installation error is related to the pyo3 library: For both `aarch64` (ARM64) and `x86_64` (AMD64) architectures, the MacOS linker is too strict to resolve the python symbols at runtime. The fix, as discussed with @nreinicke, is to modify the rust configuration file (which, for default installations, is nominally at `~/.cargo/config.toml`)

I was able to reproduce the error on my machine, and it was fixed with the changes to the config file. 

A description of the error and the config file additions are included in the new docs page.

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -30,3 +30,4 @@ parts:
       - file: developers/build_the_docs
       - file: developers/contributing
       - file: developers/rust_code_style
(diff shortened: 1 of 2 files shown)


nreinicke (approved):
This looks good, thanks for adding this!

PR #533, opened 2026-06-22
Introduce ModelIdentifier for ID options during energy model handling

### LINKED ISSUE #529: EnergyModelService supporting structured model_name entries
at present, a powertrain model name is a string, such as `2016_TOYOTA_Camry_4cyl_2WD`. this fully-qualified representation of a v1 model concatenates components of a model such as its make, model, year, and other details. in powertrain v2, fully-qualified string identifiers will still be supported, but, optionally, so will structured JSON identifiers.

both variants can be expressed as an enum:

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash)]
#[serde(untagged)]
pub enum ModelIdentifier {
  FullyQualifiedId(String),
  StructuredId {
    make: String,
    model: String,
    year: NonZeroU64,
    Variant: Option<String>,
    Version: Option<String>
  }
}
```

### integration

in order to support both types of ids, we need to modify `EnergyModelService` so that
  - the `vehicle_library` stores a `HashMap<ModelIdentifier, Arc<dyn TraversalModelService>>`
  - the `EnergyModelService::build` method must deserialize a ModelIdentifier from the value at "model_identifier", replacing "model_name"
  - deserializing `name` in `EnergyModelBuilderConfig` should target a ModelIdentifier, leaving "name" as the expected key for now

### acceptance criteria

  - if CI passes on this branch, we know the current behavior (FullyQualifiedId) still works
  - a unit test should be added with a payload containing a StructuredId which should not fail

test case for unit test:
```json
{
  "name": {
    "make": "Ford",
    "model": "Quadricycle",
    "year": 1896
  },
  "type": "ice",
  "mass_estimate_lbs": 500,
  "model_input_file": "ford_quad_1896.bin",
  "energy_rate_unit": "gallons gasoline/mile",
  "real_world_energy_adjustment": 100.0,
  "a_star_heuristic_energy_rate": 0.001,
  "input_features": [
    {
      "type": "speed",
      "name": "edge_speed",
      "unit": "mph"
    }
  ],
  "model_type": "onnx"
}
```

admrtin:
## Summary
* Closes #529 
* Added a new `ModelIdentifier` enum and updated `EnergyModelBuilder` and `EnergyModelService` to reflect the addition of structured IDs.
* `vehicle_library` is now type `HashMap<ModelIdentifier, Arc<dyn TraversalModelService>>`
* `model_name` variable in both the builder and service has changed to `model_identifier`

## Testing: 
Unit tests for the new `ModelIdentifier` class pass and workspace tests pass (not shown in image):

Screenshot 2026-06-22 at 11 41 55 AM

**Testing Note**: The unit tests mocked up the vehicle_json files with strings rather than through files I/O. If you would like me to incorporate tests for reading in the files let me know.

### DIFF
--- a/rust/routee-compass-powertrain/src/model/energy_model_service.rs
+++ b/rust/routee-compass-powertrain/src/model/energy_model_service.rs
@@ -38,15 +38,11 @@ impl TraversalModelService for EnergyModelService {
 
         let service = self.vehicle_library.get(&model_identifier).ok_or_else(|| {
             TraversalModelError::BuildError(format!(
-                "unknown vehicle model {:?}, must be one of [{}]",
-                model_identifier,
-                self.vehicle_library
-                    .keys()
-                    .map(|m| format!("{m:?}"))
-                    .collect::<Vec<_>>()
-                    .join(", ")
+                "unknown vehicle model {} from `model_name` field in query",
+                model_identifier
             ))
         })?;
+
         let model = service.build(parameters)?;
         Ok(model)
     }
--- a/rust/routee-compass-powertrain/src/model/energy_model_service.rs
+++ b/rust/routee-compass-powertrain/src/model/energy_model_service.rs
@@ -1,3 +1,4 @@
+use crate::model::model_identifier::ModelIdentifier;
 use routee_compass_core::model::traversal::{
     TraversalModel, TraversalModelError, TraversalModelService,
 };
@@ -8,12 +9,12 @@ use std::sync::Arc;
 /// based on the model_name field of the incoming query.
 #[derive(Clone)]
 pub struct EnergyModelService {
-    pub vehicle_library: HashMap<String, Arc<dyn TraversalModelService>>,
+    pub vehicle_library: HashMap<ModelIdentifier, Arc<dyn TraversalModelService>>,
 }
 
 impl EnergyModelService {
     pub fn new(
-        vehicle_library: HashMap<String, Arc<dyn TraversalModelService>>,
+        vehicle_library: HashMap<ModelIdentifier, Arc<dyn TraversalModelService>>,
     ) -> Result<Self, TraversalModelError> {
         Ok(EnergyModelService { vehicle_library })
     }
@@ -24,25 +25,22 @@ impl TraversalModelService for EnergyModelService {
         &self,
         parameters: &serde_json::Value,
     ) -> Result<Arc<dyn TraversalModel>, TraversalModelError> {
-        let model_name = parameters
-            .get("model_name")
-            .ok_or_else(|| {
-                TraversalModelError::BuildError("query missing 'model_name' field".to_string())
-            })?
-            .as_str()
-            .ok_or_else(|| {
-                TraversalModelError::BuildError("query 'model_name' is not a string".to_string())
+        let model_identifier: ModelIdentifier = serde_json::from_value(parameters.clone())
+            .map_err(|e| {
+                TraversalModelError::BuildError(format!(
+                    "Query missing 'model_identifier' field\nFrom serde_json::{e}"
+                ))
             })?;
 
-        let service = self.vehicle_library.get(model_name).ok_or_else(|| {
+        let service = self.vehicle_library.get(&model_identifier).ok_or_else(|| {
             TraversalModelError::BuildError(format!(
-                "unknown vehicle model {}, must be one of [{}]",
-                model_name,
+                "unknown vehicle model {:?}, must be one of [{}]",
+                model_identifier,
                 self.vehicle_library
                     .keys()
-                    .cloned()
+                    .map(|m| format!("{m:?}"))
                     .collect::<Vec<_>>()
-                    .join(",")
+                    .join(", ")
             ))
         })?;
         let model = service.build(parameters)?;
--- a/rust/routee-compass-powertrain/src/model/energy_model_service.rs
+++ b/rust/routee-compass-powertrain/src/model/energy_model_service.rs
@@ -25,14 +25,14 @@ impl TraversalModelService for EnergyModelService {
         &self,
         parameters: &serde_json::Value,
     ) -> Result<Arc<dyn TraversalModel>, TraversalModelError> {
-        let model_identifier_value = parameters.get("model_name").ok_or_else(|| {
-            TraversalModelError::BuildError("query missing 'model_identifier' field".to_string())
+        let model_name: &serde_json::Value = parameters.get("model_name").ok_or_else(|| {
+            TraversalModelError::BuildError("query missing 'model_name' field".to_string())
         })?;
 
-        let model_identifier: ModelIdentifier =
-            serde_json::from_value(model_identifier_value.clone()).map_err(|e| {
+        let model_identifier: ModelIdentifier = serde_json::from_value(model_name.clone())
+            .map_err(|e| {
                 TraversalModelError::BuildError(format!(
-                    "Could not deserialize 'model_identifier' field into ModelIdentifier: {e}"
+                    "Could not deserialize 'model_name' field from query into ModelIdentifier: {e}"
                 ))
             })?;
 
--- a/rust/routee-compass-powertrain/src/model/energy_model_service.rs
+++ b/rust/routee-compass-powertrain/src/model/energy_model_service.rs
@@ -25,10 +25,14 @@ impl TraversalModelService for EnergyModelService {
         &self,
         parameters: &serde_json::Value,
     ) -> Result<Arc<dyn TraversalModel>, TraversalModelError> {
-        let model_identifier: ModelIdentifier = serde_json::from_value(parameters.clone())
-            .map_err(|e| {
+        let model_identifier_value = parameters.get("model_name").ok_or_else(|| {
+            TraversalModelError::BuildError("query missing 'model_identifier' field".to_string())
+        })?;
+
+        let model_identifier: ModelIdentifier =
+            serde_json::from_value(model_identifier_value.clone()).map_err(|e| {
                 TraversalModelError::BuildError(format!(
-                    "Query missing 'model_identifier' field\nFrom serde_json::{e}"
+                    "Could not deserialize 'model_identifier' field into ModelIdentifier: {e}"
                 ))
             })?;
 
--- a/rust/routee-compass-powertrain/src/model/model_identifier.rs
+++ b/rust/routee-compass-powertrain/src/model/model_identifier.rs
@@ -3,18 +3,19 @@ use serde_json;
 use std::fmt;
 use std::fmt::Display;
 use std::num::NonZeroU64;
+use std::str::FromStr;
 
-/// The ModelIdentifier is a type deserialized from a RouteE Powertrain .json file's "name" field,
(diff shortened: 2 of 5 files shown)


rust/routee-compass-powertrain/src/model/energy_model_service.rs
```diff
@@ -24,25 +25,26 @@ impl TraversalModelService for EnergyModelService {
...
+        let service = self.vehicle_library.get(&model_identifier).ok_or_else(|| {
             TraversalModelError::BuildError(format!(
-                "unknown vehicle model {}, must be one of [{}]",
-                model_name,
+                "unknown vehicle model {:?}, must be one of [{}]",
+                model_identifier,
                 self.vehicle_library
                     .keys()
-                    .cloned()
+                    .map(|m| format!("{m:?}"))
```
robfitzgerald: now that we are working with ModelIdentifiers, this can sometimes be a structured record, which does not have a default `std::fmt::Display` implementation, hence we now need to rely on its Debug repr instead. but this will look messy, something like `ModelIdentifier::StructuredId { make: "Honda", model: "Civic", year: 2018, variant: None, version: None }`. but if we add an implementation of std::fmt::Display, we could either:
  1. deserialize `StructuredId` instances as a minified JSON via `serde_json::to_string`
  2. deserialize `StructuredId` instances as their corresponding `FullyQualifiedId` by underscore-concatenating the fields (skipping None instances)

one of these makes a much clearer error message for users, since whatever's currently printed there cannot be passed as a model `name`.

@nreinicke thoughts? @admrtin thoughts?

admrtin: @robfitzgerald @nreinicke For that error message in particular, printing out all fields of the `StructuredID` for every single model in `vehicle_library` might be too overwhelming to the user. I'm leaning towards using `FullyQualifiedID` by default, but wondering there's a way to have both. E.g., what if we introduce a new error type `ModelIdentifierError` that always prints a template example of the `StructuredID` json format vs `FullyQualifiedID` before printing out the actual object information as `FullyQualifiedID`? Is that overkill?

E.g.,

```text
ModelIdentifierError: In PATH_TO_CONFIG.json, an invalid vehicle_model was chosen.

vehicle_model can be of type:

ModelIdentifier::FullyQualifiedID

Example:

"vehicle_model" = "year_make_model"

or of type:

ModelIdentifier::StructuredID

Example:

"vehicle_model" = {
  "make" = "make",
  "model" = "model",
  "year" = 1234
},


Please choose a valid vehicle_model from the vehicle_library:
2010_Mazda_3_i-Stop
2012_Ford_Focus
2012_Ford_Fusion
2016_AUDI_A3_4cyl_2WD
...
```

robfitzgerald: this is nice thinking @admrtin. there is definitely room here to improve how we make this good UX. 

what i'm worried about, regarding both this suggestion but also the existing behavior, is that this error occurs in the `build` method of the `TraversalModelService` implementation for `EnergyModelService`. this means the incoming config values are not coming from a config file, but instead, from a _single row of RouteE Compass queries_, and the error isn't coming up at service initialization but instead _at query evaluation time_. 

so, if we provide something verbose, it will end up in every output row of a conventional compass run. for example, take a look at how overly-verbose the existing map matching error is in every output row of the current bambam run on Kestrel. i'll skip the header row for now for brevity, but you should get the picture, it's the only column that's its own James Joyce novel (scroll right):

```
null,,"881659a925fffff",,,,,,51.9757843258,179.5609131209,,,,,,,,"walk",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0.0211842905873559,"While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (179.56091, 51.975784) to map but no valid connections were found within tolerance of 0.53141401 kilometers",,,,,,,
null,,"8816593361fffff",,,,,,51.9142725892,179.6504332774,,,,,,,,"walk",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0.0211372657363572,"While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (179.65044, 51.914272) to map but no valid connections were found within tolerance of 0.53141401 kilometers",,,,,,,
null,,"881659046bfffff",,,,,,51.9173936437,179.4944444437,,,,,,,,"bike",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0.0211557370743791,"While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (179.49445, 51.917393) to map but no valid connections were found within tolerance of 0.53141401 kilometers",,,,,,,
null,,"881659a9b1fffff",,,,,,51.9476289412,179.6953085327,,,,,,,,"walk",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0.021152761501987,"While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (179.69531, 51.947628) to map but no valid connections were found within tolerance of 0.53141401 kilometers",,,,,,,
null,,"881659a99bfffff",,,,,,51.9684901762,179.7293652186,,,,,,,,"bike",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,0.0211618118636823,"While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (179.72937, 51.96849) to map but no valid connections were found within tolerance of 0.53141401 kilometers",,,,,,,
```

at present that run has a 21GB output csv (uncompressed) with 16,347,003 output rows, where 23% rows have errors. we need those errors to be clear but succinct so they don't bloat too much. we probably also need them to print nicely without newlines

happy to discuss more!

admrtin: @robfitzgerald I see! That sounds good to me, thanks for the message. I haven't really thought about error brevity from a data storage perspective, nor did I realize how many queries we might be doing. this is good insight. I think keeping it to `FullyQualifiedID` makes a lot of sense if we're maintaining the error with the `vehicle_library` printout... but judging from this discussion, **it seems like we should remove the library printout and make the error message more readable**

If y'all agree then I think I prefer the `StructuredID` as the `Display` trait.

Is there any way we could have the available `vehicle_model` and their `ModelIdentifier` variants in a web link or through the api docs of the `routee-compass` or `routee-powertrain` repository? Then we could just paste that link in the error message.

robfitzgerald: > it seems like we should remove the library printout and make the error message more readable

totally.

> If y'all agree then I think I prefer the StructuredID as the Display trait.

if you mean the serde_json::to_string of the structured value, that sounds good to me.

> Is there any way we could have the available vehicle_model and their ModelIdentifier variants in a web link or through the api docs of the routee-compass or routee-powertrain repository? Then we could just paste that link in the error message.

future work. we will have this, but we also still want to support local repositories overriding the web location. to include that information correctly in this error message, we would need to pipe down the EnergyModelBuilder configuration into the service where this error occurs. more future work.

rust/routee-compass-powertrain/src/model/model_identifier.rs
```diff
@@ -0,0 +1,309 @@
...
+/// `ModelIdentifierError` (for `TryFrom<&ModelIdentifier> for ModelIdentifier` trait implementation)
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum ModelIdentifierError {
+    #[error("fully qualified id '{0}' must have at least make/model/year fields")]
+    MissingFields(String),
+    #[error("invalid year '{0}' in fully qualified id")]
+    InvalidYear(String),
+}
+
+impl TryFrom<&ModelIdentifier> for ModelIdentifier {
```
robfitzgerald: this is built on the right fundamentals - leveraging Rust's built-in traits (TryFrom in this case) to expose discoverable APIs on types. but, semantically, this one is a bit confusing, and that's because we are not really targeting a _type_ but a _variant_:

```rust
let m1 = Modelidentifier::FullyQualifiedId("make/model/1995".to_string());
let m2 = ModelIdentifier::try_from(&m1).unwrap();
```

i would expect that this function would do nothing, since we already have a ModelIdentifier. or, i'll remember it targets one of the variants but i'll never remember which one.

a clearer convention here is to create enum methods instead:

```rust
impl ModelIdentifier {
  pub fn to_structured(&self) -> Result<Self, ModelIdentifierError> {
    match self {
      Self::StructuredId { .. } => Ok(self.clone()),
      Self::FullyQualifiedId(string) => /* attempt conversion */
    }
  }

  pub fn to_fully_qualified(&self) -> Self {
    /* this way can't fail actually */
  }
}
```

rust/routee-compass-powertrain/src/model/model_identifier.rs
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+    }
+}
+
+impl Display for ModelIdentifier {
+    /// `Display` trait `fmt()` implementation for `ModelIdentifier`.
+    /// If self is `FullyQualifiedID`, prints as `StructuredID`.
+    /// If self is `StructuredID`, prints as `StructuredID`.
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
robfitzgerald, line 136: i must have misunderstood our conversation from before:

> Adam: If y'all agree then I think I prefer the StructuredID as the Display trait.
>
> Rob: if you mean the serde_json::to_string of the structured value, that sounds good to me.

so, i think here, we want to Display using _whatever variant the user provided_:
  - if it's a structured id, do what you're doing
  - if it's a fully-qualified id, print that

this way, if things are getting debugged or logged to the user, they appear in the same format that the user provided.

rust/routee-compass-powertrain/src/model/model_identifier.rs
```diff
@@ -0,0 +1,309 @@
...
+        model: String,
+        year: NonZeroU64,
+        variant: Option<String>,
+        version: Option<String>,
+    },
+}
+
+/// `ModelIdentifierError` (for `TryFrom<&ModelIdentifier> for ModelIdentifier` trait implementation)
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum ModelIdentifierError {
```
robfitzgerald, line 48: nice addition!

robfitzgerald (approved):
pretty well sorted. thank you for providing a few test cases. there is a mix-up and i've got a request to address it, as well as a request to change the API between variants. but it's pretty close. thanks!

robfitzgerald (changes requested):
sorry, everything i just said, but, _as a request for changes_ 😊

rust/routee-compass-powertrain/src/model/model_identifier.rs
admrtin: Sounds good, can do!

rust/routee-compass-powertrain/src/model/model_identifier.rs
admrtin: That's much clearer, thank you for the feedback!!

robfitzgerald (approved):
dig it! 🎷

PR #540, opened 2026-06-23
0.19.4 workspace change breaks "default-features = false" behavior

robfitzgerald:
evidently, adding routee-compass as a workspace-level defined dependency of routee-compass-py without placing "default-features = false" at that location breaks the opt-in behavior for routee-compass-powertrain and onnx that we desire. this small fix modifies the workspace Cargo.toml to reverse that.


nreinicke (approved):
Good catch, I had overlooked that in the 0.19.4 release.

PR #542, opened 2026-06-25
sinuosity-app CLI application

### LINKED ISSUE #530: Edge Sinuosity: Import Script
In order to provide required road network segment attributes for Powertrain V2 models, we need a dataset of edge sinuosity values for each edge in the network.

### Sinuosity

Sinuosity of a linestring is defined by  $S = \frac{LineString Length}{Haversine Length}$ for any linestring, where
  - the [Haversine](https://en.wikipedia.org/wiki/Haversine_formula) function approximates the length between source and destination over the surface of the Earth   
  - $S = 1.0$ is a perfectly straight line
  - $S = \infty$ whenever the haversine length is zero (for loops)

### processing

we want a function to process sinuosity of a given geometry:

```rust
pub fn sinuosity(linestring: geo::LineString<f32>) -> Result<f32, String>
```

this function will be called in a script that iterates over all WKT LINESTRINGs in a file and produces a sinuosity value for each. output the value `f32::INFINITY` when the haversine length would be zero (a loop).

each Compass dataset on a file system includes an edges-geometries-enumerated.txt.gz file, which, for each row, contains a [WKT geometry](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry) linestring value. using the Compass utility function `routee_compass_core::util::fs::read_utils::read_raw_file`, we can read the file as a collection of Geometry values. we can iterate over the collection and perform the sinuosity function. the result is a `Vec<f64>`.

### command line interface

in order for users to generate sinuosity inputs, we want to provide a command line tool in routee_compass::bin which takes the geometry_input_file and writes to the output_file with the following arguments:

```rust
#[derive(clap::Parser, Debug)]
#[command]
pub struct SinuosityAppCli {
    #[arg(long, default = "edges-geometries-enumerated.txt.gz")]
    pub geometry_input_file: String,
    #[arg(long, default = "edges-sinuosity-enumerated.txt")]
    pub sinuosity_output_file: String,
}
```
### traversal modeling

this does not require a traversal model implementation but can instead rely on the CustomTraversalModel for injecting arbitrary f64 values for each edge.

### acceptance criteria

provide a set of unit tests calling the sinuosity function, including:
  - for a loop, expecting infinity
  - for a straight line, expecting 1.0
  - an isosceles triangle has hypotenuse 1.0 and its src/dst points are the src/dst of the linestring covering its adjacent sides. the sinuosity should be sqrt(2)

numeric assertions should be performed using not a strict equality (==) but equality within some $\epsilon$ tolerance, like `result + epsilon = expected`.

reviewer will run the CLI app and at-a-glance confirm it produces a file with matching length and the first row is correct.

admrtin:
* Fixes #530 
* Three unit tests: Straight line, loop, unit isosceles triangle
* Clips sinuosity values < 1 to 1.
* Loops result in `f32::INFINITY`
* Uses Haversine for distance/length computations.
* Tested i/o via CLI and works.
* workspace tests pass.


admrtin:
@robfitzgerald This is ready for a review!

rust/routee-compass/src/bin/sinuosity-app.rs
```diff
@@ -0,0 +1,163 @@
...
+use geo::algorithm::line_measures::{Haversine, Length};
+use geo::{Distance, LineString};
+use routee_compass_core::util::fs::read_utils::read_raw_file;
+use std::io::{Error, ErrorKind};
+use wkt::TryFromWkt;
+
+/// From a set of WKT LineStrings in a .txt or .gz file,
+/// This CLI app will compute the sinuosity of each of
+/// the linestrings.
+
```
robfitzgerald: Remove this gap so docstring is detected by cargo docs

robfitzgerald (approved):
looks good! thanks @admrtin :trumpet: one small edit requested

Issue #351, opened 2025-07-02
Generalize the road class frontier model


robfitzgerald:
The road class frontier model is useful for the general case of filtering edges by categorical attributes. However, only one instance of this plugin can exist at-a-time in compass, as it enforces that accepted categories are listed at the "road_class" field of the incoming query.

In order to allow multiple instances of this model, its expected key should be parameterized in configuration such that the current implementation might be built with:

```toml
[[frontier.models]]
type = "categorical"
key = "road_class"
input_file = "/Users/rfitzger/dev/nrel/routee/routee-compass-tomtom/data/tomtom_denver/edges-routing-class-enumerated.txt.gz"
```

Towards this,
  - we should rename it to something like CategoricalFrontierModel
  - along the way, move the code into routee-compass-core (default frontier models should live in the core repo)

Related: #208

robfitzgerald:
as a result, we could test for two different types of categoricals on each edge. take for example this chill walking-to-a-nice-place-to-hang-out search:

in config.toml file:

```toml
[[constraint.models]]
type = "categorical"
key = "road_class"
input_file = "edges-routing-class-enumerated.txt.gz"

[[constraint.models]]
type = "categorical"
key = "destination"
input_file = "edges-destinations-enumerated.txt.gz"
```

on a given routing query:

```json
{
  "road_class": ["footpath", "sidewalk", "staircase", "service", "residential"],
  "destination": ["coffee_shop", "grocery_store", "bar", "restaurant"]
}
```

robfitzgerald:
Closed via #547

PR #547, opened 2026-07-07
Remove the `road_class` constraint type and generalize it to `categorical` constraint model

admrtin:
# Fixed Issues:
* #351 

# Summary of changes:
* This removes the `RoadClassModel` and it's associated builders and services, and instead refactors it into a more generalizable`CategoricalModel`.
* A new `key` argument was introduced for this type of constraint model, which specifies the category name.
* The `CategoricalModelService` contains the key information.


rust/routee-compass-core/src/model/constraint/default/categorical/categorical_builder_config.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+/// The configuration arguments to be passed into the CategoricalModelBuilder.
+pub struct CategoricalModelBuilderConfig {
+    pub key: String, // The type of categorical constraint service (e.g., "road class") to build
+    pub input_file: String, // The path to the input file for the constraint service
```
robfitzgerald: revert this to "road_class_input_file"

rust/routee-compass-core/src/model/constraint/default/categorical/categorical_builder.rs
```diff
@@ -0,0 +1,110 @@
...
+/// key = "destination"
+/// input_file = "edges-destinations-enumerated.txt.gz"
+/// ```
+///
+/// At query time, the search can be constrained to a subset of a category's values.
+/// For example, with the `road_class` model configured above:
+///
+/// ```json
+/// {
+///     "road_classes": ["footpath", "sidewalk", "staircase", "service", "residential"]
```
robfitzgerald: here, the key should match as above, which would make it "road_class"

rust/routee-compass-core/src/model/constraint/default/categorical/categorical_builder.rs
```diff
@@ -0,0 +1,110 @@
...
+        // since we have a new hashmap, first ID will match to None, so we incrementally build out
+        // the hashmap.
+        for class in category_lookup.iter() {
+            let id = match mapping.get(class) {
+                Some(id) => *id,
+                None => {
+                    let id_usize = next_id;
+                    if id_usize > u8::MAX as usize {
+                        return Err(ConstraintModelError::BuildError(
+                            "too many unique road classes, max is 256".to_string(),
```
robfitzgerald: replace "road classes" here with something like "{key} categories"

rust/routee-compass-core/src/model/constraint/default/categorical/categorical_model.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn valid_edge(&self, edge: &Edge) -> Result<bool, ConstraintModelError> {
+        match &self.query_categories {
+            None => Ok(true),
+            Some(encoding) => self
+                .service
+                .category_by_edge
+                .get(edge.edge_id.0)
+                .ok_or_else(|| {
+                    ConstraintModelError::ConstraintModelError(format!(
+                        "edge id {} missing from constraint model file",
```
robfitzgerald: "...missing from categorical constraint model file for category key {key}"

rust/routee-compass-core/src/model/constraint/default/categorical/categorical_service.rs
```diff
@@ -1,33 +1,44 @@
...
-            Some(Ok(road_classes)) => {
-                let mapped: Result<HashSet<u8>, ConstraintModelError> = road_classes
+            Some(Ok(categories)) => {
+                let mapped: Result<HashSet<u8>, ConstraintModelError> = categories
                     .iter()
                     .map(|c| {
-                        self.road_class_mapping.get(c).copied().ok_or_else(|| {
+                        self.category_mapping.get(c).copied().ok_or_else(|| {
                             ConstraintModelError::BuildError(format!(
                                 "road class '{}' not found in road class mapping",
```
robfitzgerald: similarly replace "road class" here with "{key} category" here

robfitzgerald (changes requested):
thanks for getting this moving. i have a few requests for small changes but in general the new functionality works as expected.

to test, i made the following modifications to the osm_default_distance.toml configuration file:

```toml
# add this missing objective function weight
[cost.weights]
trip_distance = 1.0

# add this to demo the new constraint model
[search.constraint]
type = "categorical"
key = "my_super_awesome_key"
input_file = "edges-road-class-enumerated.txt.gz"

# add this to write the result to a file
[system.response_output_policy]
type = "file"
filename = "result.json"
format = { type = "json", newline_delimited = true }
```

and ran this query, which restricted the search to only the roadway arterials.

```json
{
    "destination_y": 39.757431,
    "destination_x": -104.988567,
    "origin_y": 39.749513,
    "origin_x": -105.000316,
    "my_super_awesome_key": ["primary", "primary_link", "motorway", "motorway_link"]
}
```

clearly it worked. this is the resulting distance-minimal trip had to drive pretty far out of its way along I-70 + I-25 to reach its destination :laughing:

Image

rust/routee-compass-core/src/model/constraint/default/categorical/categorical_builder_config.rs
admrtin: hmm, can I ask why? it feels like if we are removing `road_class` from every other place in the previous version then we should also remove it here, or name it something like "categorical_input_file"

robfitzgerald: Because of the error variant that bubbles up from a utility function called normalize_file_paths which finds keys that terminate with "input_file" and confirms that the file exists. When the file does not exist, it will error with the key name, but that is not very informative when it is just "input_file". So our convention right now is to include _what kind of input file_ in the name of any input file.

robfitzgerald: Oh! Sorry, yes, let's rename it "categorical_input_file" or something like that, I missed your point!

robfitzgerald (approved):
🎺  LGTM! 🎺

PR #548, opened 2026-07-08
Update hpc build

nreinicke:
Updates our HPC build process slightly to simplify things. Previously we required custom features in our routee-compass-powertrain crate to enable static linking against the pre-compiled ONNX runtime. This removes those by just pointing ort to the pre-compiled binary and allowing it to do dynamic linking. This also updates our pixi configuration so HPC installs can be done simple with `pixi run build_hpc` and then any HPC code can run in the hpc environment (`pixi run -e hpc`).

Also bumps our github rust clippy checks to match the strict checks that we were doing locally with pixi.


robfitzgerald:
@nreinicke saw you were still making changes after pinging me. is this now ready for review?

nreinicke:
@robfitzgerald - yep, thanks for checking, this is ready

rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+[toolchain]
+# Pinned to match the `rust` version in pyproject.toml's pixi environment so that
+# `pixi run check_rust`, local `cargo`, and CI all use the same clippy/rustc and
+# catch the same lints. Bump this and the pyproject.toml pin together.
+channel = "1.96.1"
```
robfitzgerald, line 5: but this means we need to keep this number updated. perhaps we add a CI test that compares this version against the current latest rust version via rustup?

robfitzgerald (commented):
this is great, it's nice and modern tooling to get around a complex build dependency, and the docs are a nice touch too.

i'm noticing that this breaks the current "optional feature" workflows for compiling routee-compass or routee-compass-py without the "ort" dependency. this was in place for compass on HPC, but also for compiling bambam. is it possible for this build tooling to support bambam compilation on HPC? otherwise, can we roll this back so users can still compile compass without powertrain/ort?

based on your reply i'll come back around and finish the review. thanks!

nreinicke:
> i'm noticing that this breaks the current "optional feature" workflows for compiling routee-compass or routee-compass-py without the "ort" dependency. this was in place for compass on HPC, but also for compiling bambam. is it possible for this build tooling to support bambam compilation on HPC? otherwise, can we roll this back so users can still compile compass without powertrain/ort?

I think we have always been compiling ort into all downstream libraries but before this, we had the `ort-static` feature that would statically link the pre-compiled ONNX runtime to support our old HPC linux OS. Now, with these changes, the compilation should work for both regular builds and HPC builds without having to change any feature flags which should make it simpler for downstream users. routee-compass-powertrain still remains optional if you want to exclude it.

robfitzgerald:
i'll try and point a new branch on bambam at this branch of compass and see if i can get a successful build. that's the only thing i'm concerned about, since i had a few unexpected outcomes dealing with how workspace Cargo.tomls and default features interact. if that passes, then i'll accept the build tooling + additional clippy-related fixes. :fish:

> Now, with these changes, the compilation should work for both regular builds and HPC builds

do you think the [installation](https://natlabrockies.github.io/routee-compass/installation.html) documentation should get updated too? maybe this pixi-based workflow should be added to that list, under "from source" as a means to build dependencies from source as well. and while i'm mentioning it, this new set of pixi run steps _aren't actually an HPC workflow_, it's more like another "from source build" alternative toolchain, right? but one that goes a step further and builds the ONNX runtime dependency from source as well.

robfitzgerald (approved):
looks good, thank you for coming up with this workaround for our Kestrel deployments. :trumpet:

robfitzgerald:
another thought. this PR takes another step towards making pixi the defacto build tooling for contributing to routee-compass. but it's still not documented as such. if you could contribute an update to developer documentation describing how pixi is used during development, i would support that and get onboard.

nreinicke:
> his new set of pixi run steps aren't actually an HPC workflow, it's more like another "from source build" alternative toolchain, right? but one that goes a step further and builds the ONNX runtime dependency from source as well.

> if you could contribute an update to developer documentation describing how pixi is used during development, i would support that and get onboard.

Yes, these are both good points. I will take an action to update the docs to clarify and focus more on how pixi is being used as the "recommended" developer build tool.

rust-toolchain.toml
nreinicke: Ah right, that's a good point; I'll take a look at this to see how we might better keep this updated.

PR #553, opened 2026-07-10
eval plugin improvements

### LINKED ISSUE #546: eval plugin will break CSV invariants when writing an error log
the eval plugin has an argument dictating what to do when evaluation leads to an error:

```rust
#[derive(Serialize, Deserialize, Debug, Clone)]
#[serde(rename_all = "snake_case", tag = "type")]
pub enum OnFailureBehavior {
    /// interrupt the plugin on failure, returning an Err from the plugin.
    Interrupt,
    /// record the error to the output row at some JSONpath
    Record { path: String },
    /// ignore the error
    Ignore,
}
```

when recording to some JSONPath, the array of failures is serialized, but it is not properly enquoted + escaped when written to a CSV field. take for instance the bambam error evaluation row below. it has occurrences of both " and ' within the serialized array but the total array value is not enquoted. to fix these, we would need to ensure that double-quotes are escaped or there is exactly two double-quotes per serialized error output.

### sensible limits

to address cases like this where many sequential expressions are run but we are only interested in the top-k failures for debugging, it would be helpful if the Record variant had a `limit: Option<u64>`, defaulting to _no limit_, which would allow the user to specify a debug output array size limit. if a limit is applied, the serialized output should be terminated with an ellipsis if additional errors are elided (`...]`).

```
[{"expr":"n_star","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_star' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"o_ijkt * (n_star / n_j) * (f_j / f_sum)","error":"failure running plugin: JSONPath query for input 'n_j' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"f + e + h + j + r + s","error":"failure running plugin: JSONPath query for input 's' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"f + e + h + j + r + s","error":"failure running plugin: JSONPath query for input 's' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"f + e + h + j + r + s","error":"failure running plugin: JSONPath query for input 's' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"f + e + h + j + r + s","error":"failure running plugin: JSONPath query for input 'e' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik40' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik30' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik40' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik30' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik40' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik40' did not return exactly one result: nodelist expected to contain one entry, but is empty"},{"expr":"(o_ik10 * exp_mi10) + ((o_ik20 - o_ik10) * exp_mi20) + ((o_ik30 - o_ik20) * exp_mi30) + ((o_ik40 - o_ik30) * exp_mi40)","error":"failure running plugin: JSONPath query for input 'o_ik20' did not return exactly one result: nodelist expected to contain one entry, but is empty"}],,,
```

### LINKED ISSUE #552: behavior on NaN in eval plugin
the output of a BAMBAM row contains NaN MEP scores because the row because the BAMBAM code erroneously placed zero-valued opportunity totals which are the denominator of terms in the MEP calculation. an (abridged) row example:

```python
{
  'grid_id': '8826cbd1bbfffff', 
  'isochrone_10': '01030000000100000000000000', 
  'lat': 32.69065927270409, 
  'lon': -97.22701446016424, 
  'mep': nan,   # <- NaN score
  'mep_entertainment': nan,  # <- NaN score
  'opps_entertainment_10': 0.0,   # <- source of Div0 error
  'routee_error': 'While interacting with the map model outside of the context of search, an error occurred. Source: failure matching query to map: attempted to match query origin coordinate (-97.22701, 32.69066) to map but no valid connections were found within tolerance of 0.5 kilometers', 
  'runtime_input': 0.012328894, 
}
```

while this is a BAMBAM problem, it should also be possible to opt into other behaviors, like zeroing out NaNs or failing fast on NaN.

robfitzgerald:
This PR adds a few requested improvements to the eval output plugin. 

Along the way,
  - tests were added to verify each new feature
  - the eval plugin module was refactored for readability. in particular, record types stored in the ops.rs file were moved into their own separate files and two functions were refactored as TryFrom implementations.

Closes #552.
Closes #546.

### NaN evalulations

NaN is a valid f64 value. when an expression runs that divides by zero, for instance, the result can be NaN. if a single NaN occurs in a string of expressions, the NaN will bleed into all downstream expressions. this may not be desirable, and for those cases, the user can now specify a behavior for handling NaN values. if the `on_nan` config key is not provided, the default behavior "allow" (aka ignore) is used, which matches the current behavior.

```toml
[[plugin.output_plugins]]
type = "eval"
expressions = []
on_failure.type = "ignore"
on_nan.type = "allow" # also "zero", "error"
```

the enum describes their behaviors:

```rust
pub enum NotANumberBehavior {
    /// allow NaNs to get written
    #[default]
    Allow,
    /// zero out NaN values when they are the result of an expression
    Zero,
    /// error out if we return a NaN
    Error,
}
```

### error output limit

when the user has chosen the `OnFailureBehavior::Record`, the plugin will write all errors to a JSONPath on the output object. this can be problematic when the list of expressions is very long. to address this, a `limit: Option<usize>` was added to the arguments of `OnFailureBehavior::Record`. when Some limit $k$ is provided, only the _first_ $k$ errors are returned. when the key is not provided, no limit is applied.

config example only allowing up to 2 errors reported to the key `eval_error` on the output object.

```toml
[[plugin.output_plugins]]
type = "eval"
expressions = []
on_failure = { type = "record", path = "eval_error", limit = 2 }
on_nan.type = "allow"
```

### errors serialized as a JSON String

since we may map the output JSON to a CSV file in the response output policy, it is possible to write the eval_errors as an array to a CSV cell. however, if the error array contains at least two error values, then it will break CSV invariants as commas will appear in the cell. to address this, an argument `as_string` was added to `OnFailureBehavior::Record` which stringifies the JSON array to the record path.

config example:

```toml
[[plugin.output_plugins]]
type = "eval"
expressions = []
on_failure = { type = "record", path = "eval_error", as_string = true }
on_nan.type = "allow"
```


rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -1,237 +1,39 @@
...
+// /// Compiled form of [`OnFailureBehavior`] — the `Record` variant's path is
+// /// pre-parsed once at plugin construction time.
+// pub enum CompiledOnFailure {
+//     Interrupt,
+//     Record {
+//         segments: Vec<PathSegment>,
+//         limit: Option<usize>,
+//     },
+//     Ignore,
+// }
```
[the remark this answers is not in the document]
robfitzgerald: fixed

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -275,7 +77,7 @@ pub fn record_error(
             })?;
             if *i < arr.len() {
                 if let Some(slot_arr) = arr[*i].as_array_mut() {
-                    slot_arr.push(entry);
+                    insert_within_limit(slot_arr, entry, limit);
                 } else {
                     arr[*i] = serde_json::Value::Array(vec![entry]);
                 }
```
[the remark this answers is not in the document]
robfitzgerald: a value of 0 is not a sensible limit for a record behavior. we now fail at configuration time if the user passes a zero.

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -397,11 +209,12 @@ pub fn eval_and_write(
         )));
     }
     let result = eval_result?;
+    let result_cleaned = nan_behavior.apply(result)?;
 
     // 3. Write the result back into the output JSON.
-    let number = serde_json::Number::from_f64(result).ok_or_else(|| {
+    let number = serde_json::Number::from_f64(result_cleaned).ok_or_else(|| {
         OutputPluginError::OutputPluginFailed(format!(
```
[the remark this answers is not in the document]
robfitzgerald: not true:

```rust
#[cfg(test)]
mod test {
    #[test]
    fn test() {
        let value = serde_json::json!(f64::NAN);
        println!("{}", serde_json::to_string(&value).unwrap_or_default())
    }
}
```
```
running 1 test
null
test plugin::output::default::eval::compiled_config::test::test ... ok
```

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -290,6 +92,15 @@ pub fn record_error(
     Ok(())
 }
 
+/// helper function to insert a value into a vector of values if it is within an
+/// optional size limit argument.
+fn insert_within_limit(arr: &mut Vec<Value>, entry: Value, limit: Option<usize>) {
+    match limit {
+        Some(lim) if arr.len() >= lim => {}
+        _ => arr.push(entry),
+    }
```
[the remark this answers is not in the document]
robfitzgerald: in order to do this in a way that is valid JSON, i will append "...and {n} more" as a value in the error list.

rust/routee-compass/src/plugin/output/default/eval/config.rs
```diff
@@ -2,13 +2,19 @@ use std::collections::HashMap;
...
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct EvalOutputPluginConfig {
     /// expressions to run for each row
     pub expressions: Vec<ExpressionConfig>,
     /// behavior when expression fails
     pub on_failure: OnFailureBehavior,
+    /// behavior when a NaN value is produced by an expression. by default,
+    /// NaN values are allowed.
+    #[serde(default)]
+    pub on_nan: NotANumberBehavior,
```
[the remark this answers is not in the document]
robfitzgerald: that's a nit, but i updated the PR comment

rust/routee-compass/src/plugin/output/default/eval/compiled_config.rs
```diff
@@ -0,0 +1,120 @@
...
+                crate::plugin::PluginError::BuildFailed(format!(
+                    "failed to parse expression '{}': {e}",
+                    conf.expr
+                ))
+            })?;
+        let compiled = parsed.from(&slab.ps).compile(&slab.ps, &mut slab.cs);
+
+        Ok(CompiledExpression {
+            inputs,
+            expr: conf.expr.clone(),
```
admrtin, line 79: nit: is clone needed here?

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -412,8 +219,144 @@ pub fn eval_and_write(
...
+/// The leading `$.` or `$` prefix is stripped before parsing.
+pub fn parse_path_segments(path: &str) -> Result<Vec<PathSegment>, OutputPluginError> {
+    let stripped = path
+        .strip_prefix("$.")
+        .or_else(|| path.strip_prefix('$'))
+        .unwrap_or(path);
+
+    if stripped.is_empty() {
+        return Err(OutputPluginError::OutputPluginFailed(
+            "output path must not be empty after '$'".to_string(),
```
admrtin, line 300: nit: maybe inject the path object itself into this error message so that the user knows what value to fix

rust/routee-compass/src/plugin/output/default/eval/config.rs
```diff
@@ -50,11 +56,35 @@ pub enum OnFailureBehavior {
...
+        #[serde(default)]
+        as_string: bool,
+    },
     /// ignore the error
     Ignore,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub enum NotANumberBehavior {
```
admrtin, line 78: I like these variants! very clear how NaNs should be handled by the plugin.

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -412,8 +219,144 @@ pub fn eval_and_write(
...
+
+/// Parse a JSONPath-style string into segments for writing.
+///
+/// Supports dot notation (`$.a.b`) and bracket array indices (`$.a[0].b`).
+/// The leading `$.` or `$` prefix is stripped before parsing.
+pub fn parse_path_segments(path: &str) -> Result<Vec<PathSegment>, OutputPluginError> {
+    let stripped = path
+        .strip_prefix("$.")
+        .or_else(|| path.strip_prefix('$'))
+        .unwrap_or(path);
```
admrtin, line 296: what happens if the path does not have $ or $. as the prefix? is this already considered?

rust/routee-compass/src/plugin/output/default/eval/ops.rs
```diff
@@ -412,8 +219,144 @@ pub fn eval_and_write(
...
 }
 
+/// Recursively walk `root` along `segments` and write `value` at the final location.
+/// Intermediate objects that do not exist are created automatically.
+pub fn set_path(
+    root: &mut serde_json::Value,
+    segments: &[PathSegment],
+    value: serde_json::Value,
+) -> Result<(), OutputPluginError> {
+    match segments {
```
admrtin, line 229: nit: i spent some time trying to understand this (still don't quite understand, maybe because of my lack of knowledge of Json trees/arrays) and I think it might be useful to add one liner comments for each of the match patterns

rust/routee-compass/src/plugin/output/default/eval/plugin.rs
```diff
@@ -1,13 +1,14 @@
 use crate::plugin::output::{
-    default::eval::ops::{self, CompiledExpression, CompiledOnFailure},
+    default::eval::{ops, CompiledExpression, CompiledOnFailure, NotANumberBehavior},
     OutputPlugin,
 };
 
-use super::config::{EvalOutputPluginConfig, OnFailureBehavior};
+use super::config::EvalOutputPluginConfig;
 
 pub struct EvalOutputPlugin {
```
admrtin, line 15: Recommend a doc comment on this struct

admrtin (changes requested):
@robfitzgerald It looks good! most of my comments are nits, readability changes, and questions about Rust syntax/the code.

rust/routee-compass/src/plugin/output/default/eval/compiled_config.rs
robfitzgerald: yes, since the value conf is a reference, transitively, so is conf.expr. two thoughts:
  - we could change this into `TryFrom<ExpressionConfig>` and omit the reference, where we wouldn't need the clone, but then the config would be dropped at the end of this call, and then the config couldn't be retained by the longer-running plugin if we wanted to. example: if we called this function and it failed, and when writing the error in the outer scope, we wouldn't be able to include anything about the config that failed because it would already be dropped.
  - this clone is small and happens exactly once during the init phase. there's a lot of this going on in compass and bambam, of really small config values getting cloned during `build` methods, and it's really not a performance hit.

rust/routee-compass/src/plugin/output/default/eval/ops.rs
robfitzgerald: if `stripped.is_empty()` then that means the path is exactly made up of _at least one_ "$." or "$" (could be "$.$.$.$." i guess but then i feel like we're getting trolled). so, wouldn't injecting the path object be redundant?

rust/routee-compass/src/plugin/output/default/eval/ops.rs
robfitzgerald: that would be fine. basically, i'm trying to map JSONPath to python-like object paths here. if $ or $. are not there, then all that's left is what we want.

rust/routee-compass/src/plugin/output/default/eval/ops.rs
robfitzgerald: i just did this, but, i find the comments are redundant, so i'm not going to commit the change. here are the comments:

```rust
pub fn set_path(
    root: &mut serde_json::Value,
    segments: &[PathSegment],
    value: serde_json::Value,
) -> Result<(), OutputPluginError> {
    match segments {
        // base case: no path segments provided
        [] => ...,
        // a single path segment that is an object key
        [PathSegment::Key(k)] => ...
        // a single path segment that is an array index
        [PathSegment::Index(i)] => ...
        // a single path segment that is an object key, followed by the rest (recurses on rest)
        [PathSegment::Key(k), rest @ ..] => ...
        // a single path segment that is an object key, followed by the rest (recurses on rest)
        [PathSegment::Index(i), rest @ ..] => ...
    }
}
```

edit: happy to talk through this if it's still unclear. it's a combination of slice pattern matching styles with recursive algorithms

rust/routee-compass/src/plugin/output/default/eval/plugin.rs
robfitzgerald: done!

rust/routee-compass/src/plugin/output/default/eval/ops.rs
admrtin: that makes sense! thanks.

rust/routee-compass/src/plugin/output/default/eval/ops.rs
admrtin: thank you!

admrtin (approved):
Good to go!

PR #555, opened 2026-07-20
Trip history traversal model

### LINKED ISSUE #527: TripHistoryTraversalModel
in RouteE Powertrain V2, models may leverage the values of previous links in order to estimate energy consumption. in Compass, this could be found in one of two ways: storing/retrieving in the state vector, or, traversing the search tree at each link to find the values. these have a few clear tradeoffs. if we store the values in the state vector, it increases our memory requirements substantially. if we traverse the tree at each step, then we use a lot more CPU to collect the values.

in Compass, we prefer increases in RAM to increases in runtime. additionally, storing the values in the state vector gives us observability when reviewing the results.

a utility for storing and retrieving previous link values is a TraversalModel:

```rust
pub struct TripHistoryModelConfig {
  /// the features on the present link that will be collected in the history. these are the values
  /// collected by other traversal models stored into the state vector that will be `shifted` into
  /// slots representing past links as the links are traversed
  input_features: Vec<InputFeature>,
  /// how deep into the past link attributes will be collected. 
  depth: std::num::NonZeroUsize
}

pub struct TripHistoryService {
  config: TripHistoryModelConfig,
  engine: Arc<TripHistoryModelEngine>,
}

pub struct TripHistoryModel {
  engine: Arc<TripHistoryModelEngine>
}

impl TraversalModel for TripHistoryModel {
  ...
}
```

this utility would:
  - create named state vector fields with normalized names taken from `format!("{feature_name}_{depth_value}")` as the cross product of `features` and `(0+1)..depth`. zero is skipped as this value is encoded without an enumeration (i.e., "energy_liquid" is the zeroth entry, "energy_liquid_1" is the previous link value
  - ~create a named state vector feature `depth: u64` to track the depth stored in the current vector (used to encode which fields are "empty" when tree depth < the history depth)~
  - use type-specific sentinel values for tracking whether a value has been stored at some depth (f64::NAN, i64::-1, u64::MAX, bool::false) and set these as the initial values of the `StateVariableConfig` entries for the history slots. those are stored on the Engine and returned by the Model via `TraversalModel::output_features`.
  - provide a `fn shift(&mut [StateVariable])` function on the Engine that takes all values from depth values `1..(depth)` and _shifts_ the value from `format!("{feature_name}_{depth_value}")` to `format!("{feature_name}_{depth_value+1}")` in the state vector, shifting values one link into "the past".
  - provide a `fn insert_first()` function on the Engine to traverse **one step** into the history via `ctx.tree.backtrack_with_depth(state_variable, depth)` and record the value at `format!("{feature_name}_1")`. this must be run after `shift` in order to avoid overwriting the 1st history value before shifting. if the backtrack result is empty, do nothing.
  - provide a `pub fn update_history()` function that calls `shift` followed by `insert_first`
  - in the `TraversalModel::traverse_edge` function, call `shift` and finish.
  - in the `TraversalModel::estimate_traversal` function, do nothing -> `Ok(())`

this utility does not need to know how to insert the features into the current link (i.e., "energy_liquid").

### acceptance criteria

- [x] this model is added to the routee-compass-core crate, following the pattern of existing traversal models
- [x] Builder/Service/Model implemented
- [x] builder wired into the compass_builder_inventory.rs in the routee-compass crate
- [x] provide unit tests for `shift` to demonstrate the expected behavior up to 5 links. this will require mocking with a StateModel and at least one other upstream model such as DistanceTraversalModel.
- [x] provide unit tests for `insert_first` to demonstrate the expected behavior for an empty tree and a tree with one branch

admrtin:
New traversal model which is meant to close issue #527 

The `TripHistoryTraversalModel` is a new traversal model that allows users to specify what `input_features` to grab at a certain `depth` in the previous state vectors. These historical features (hence, "trip history") will be used to predict energy in the new Powertrain V2 models.

TODOs:
- [x] Switch the `shift` method to name-based indexing
- [x] Switch the `insert_first` method to name-based indexing
- [x] Rewrite unit test for `shift`
- [x] Write unit test for `insert_first`


admrtin:
@robfitzgerald I'm not sure why, but `ruff check` keeps failing in the CI, but it appears fine on my machine. 

I think it's because the PR is failing on pre-existing Python Ruff violations from a new CI test suite, not on changes introduced here. Did we update the CI tests recently?

Not sure if I should address these lints here or in a separate PR.

robfitzgerald:
@admrtin @nreinicke [this](https://astral.sh/blog/ruff-v0.16.0) is likely the explanation:

> Ruff now enables 413 rules by default, up from 59 in previous versions.

our CI doesn't tag ruff to a release version so CI started running the latest as soon as it was released and it caught a bunch of new rules.

i'll create an issue (here: #559) to bump ruff to 0.16.0 and deal with all of the linter errors. can you update the pyproject.toml file in your PR so that ruff 0.15.22 is installed.

```toml
[project.optional-dependencies]
dev = [
  ...
  "ruff==0.15.22",
```

alternatively, we could leave the pyproject file as is and instead add a .ruff.toml file (or pyproject ruff section) with this suggestion from the blog post:

> If you want to revert to the old default set, you can easily select the old rules with this configuration (here adapted for pyproject.toml file):

```toml
[tool.ruff.lint]
select = ["E4", "E7", "E9", "F"]
```

admrtin:
@robfitzgerald can do! thanks for the help here!

admrtin:
@robfitzgerald The tests are passing now so it's ready for a review. I can fix the linter errors in a separate PR after we get this PR to main.

rust/routee-compass-core/src/model/state/state_variable_config.rs
```diff
@@ -11,38 +11,52 @@ use uom::si::f64::*;
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum StateVariableConfig {
     Distance {
+        #[serde(default)]
```
robfitzgerald: @nreinicke Adam has this suggestion to allow configurations to omit `initial` and `accumulator` values, defaulting to 0.0 and false respectively. this cuts down on verbosity but could maybe confuse someone who doesn't know those defaults. something to discuss in our weekly.

rust/routee-compass-core/src/model/traversal/default/trip_history/trip_history_traversal_model.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn name(&self) -> String {
+        "TripHistoryModel".to_string()
+    }
+
+    fn input_features(&self) -> Vec<InputFeature> {
+        // the set of input features from the trip history configuration
+        self.engine
+            .history_features
+            .iter()
+            .map(|feature| match &feature.state_variable_config {
```
robfitzgerald: figure we would want to use the new From (or from_state_variable_config) implementation here that you have created above.

rust/routee-compass-core/src/model/traversal/default/trip_history/trip_history_traversal_model.rs
```diff
@@ -0,0 +1,145 @@
...
+        _od: (&Vertex, &Vertex),
+        _state: &mut Vec<StateVariable>,
+        _tree: &SearchTree,
+        _state_model: &StateModel,
+    ) -> Result<(), TraversalModelError> {
+        Ok(())
+    }
+}
+
+fn set_initial_as_sentinel(mut conf: StateVariableConfig) -> StateVariableConfig {
```
robfitzgerald, line 96: doc comment for this function explaining what the sentinel is meant to represent (in this case, "unset")

rust/routee-compass-core/src/model/state/input_feature.rs
```diff
@@ -54,6 +57,53 @@ impl InputFeature {
     }
 }
 
+impl From<&StateVariableConfig> for InputFeature {
+    fn from(value: &StateVariableConfig) -> Self {
+        match value {
+            StateVariableConfig::Distance { .. } => InputFeature::Distance {
```
robfitzgerald, line 63: i do like the idea of this but i don't believe it is implemented correctly. the state variable config does not store the InputFeature::name value. so perhaps instead we need

```rust
impl InputFeature {
  pub fn from_state_variable_config(fieldname: &str, config: &StateVariableConfig) -> InputFeature {
    todo!()
  }
}
```

rust/routee-compass-core/src/model/traversal/default/trip_history/trip_history_traversal_engine.rs
```diff
@@ -0,0 +1,534 @@
...
+            &dst_vertex,
+            &mock_edge,
+            &parent_label,
+        );
+
+        // update the trip history: shift -> insert_first
+        trip_history_engine
+            .update_history(&ctx, &mut state, &state_model)
+            .unwrap();
+
```
robfitzgerald, line 522: ~should we also assert here that slot 0 ("edge_distance") is unchanged? i debugged the output (using state_model.serialize_state) and found that it wasn't:~

oop, i get it, that's the value that is copied from the tree.

robfitzgerald (changes requested):
looks good! the history algorithms appear correct, thank you for the test implementations. there is a perhaps un-needed From implementation we could do away with or modify, i leave it to you, and that's the only change i'm eyeballing here. let's leave the determination on the #[serde(default)] macros for tomorrow's meeting. thanks Adam!

admrtin:
@robfitzgerald I made the requested changes!

rust/routee-compass-core/src/model/state/state_variable_config.rs
robfitzgerald: maybe we should roll these #[serde(default)] macros because both Nick and I can't fully convince ourselves it's the right solution for this today, and i don't want to hold up your PR for too long.

admrtin: That sounds good to me! ~~will just need to make some minor tweaks to the unit tests and remove the defaults, then.~~

admrtin:
@robfitzgerald the `StateVariableConfig` default changes are rolled back.

robfitzgerald (approved):
🥇

PR #562, opened 2026-07-29
Bumps ruff requirement to >= 0.16.0 to match CI tests and fixes all ruff check errors

### LINKED ISSUE #559: Use ruff 0.16
As noted in the [astral blog entry](https://astral.sh/blog/ruff-v0.16.0) for ruff 0.16.0:

> Ruff now enables 413 rules by default, up from 59 in previous versions.

When using ruff 0.16.0, many linter errors are generated which must be addressed so that the CI linter test will pass.

admrtin:
Fixes #559.

@robfitzgerald @nreinicke I wanted to mention that there were some generic `Exception` types that ruff was mad about in some of the examples and in the pipeline, so I changed them to `RuntimeError`, but might be worth a sanity check. I marked all the ones I thought you may want to take a look at in the conversation below.

### DIFF
--- a/python/nrel/routee/compass/io/charging_station_ops.py
+++ b/python/nrel/routee/compass/io/charging_station_ops.py
@@ -1,13 +1,13 @@
 import importlib.resources
-import requests
-import re
-from typing import List, Union, TYPE_CHECKING
-
 import logging
+import re
+from typing import TYPE_CHECKING, Union
+
+import requests
 
 if TYPE_CHECKING:
-    from pandas import DataFrame
     from geopandas import GeoDataFrame
+    from pandas import DataFrame
     from shapely.geometry import Polygon
 
 log = logging.getLogger(__name__)
@@ -43,10 +43,10 @@ def download_ev_charging_stations(state: str, api_key: str = "DEMO_KEY") -> "Dat
         response.raise_for_status()
         result = response.json()
     except requests.RequestException as e:
-        raise Exception(f"Failed to download charging station data: {e}")
+        raise RuntimeError(f"Failed to download charging station data: {e}")
 
     if "fuel_stations" not in result:
-        raise Exception("Invalid response format from NREL API")
+        raise RuntimeError("Invalid response format from NREL API")
 
     # Create dataframe
     df = pd.DataFrame(result["fuel_stations"])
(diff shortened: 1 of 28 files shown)


python/nrel/routee/compass/io/charging_station_ops.py
```diff
@@ -43,10 +43,10 @@ def download_ev_charging_stations(state: str, api_key: str = "DEMO_KEY") -> "Dat
         response.raise_for_status()
         result = response.json()
     except requests.RequestException as e:
-        raise Exception(f"Failed to download charging station data: {e}")
+        raise RuntimeError(f"Failed to download charging station data: {e}")
```
admrtin, line 46: exception change here

python/nrel/routee/compass/io/charging_station_ops.py
```diff
@@ -43,10 +43,10 @@ def download_ev_charging_stations(state: str, api_key: str = "DEMO_KEY") -> "Dat
...
 
     if "fuel_stations" not in result:
-        raise Exception("Invalid response format from NREL API")
+        raise RuntimeError("Invalid response format from NREL API")
```
admrtin, line 49: exception change here

python/nrel/routee/compass/io/charging_station_ops.py
```diff
@@ -340,7 +340,7 @@ def download_ev_charging_stations_for_polygon(
                 # Add state column for reference
                 state_stations["state"] = state
                 all_stations.append(state_stations)
-        except Exception as e:
+        except RuntimeError as e:
```
admrtin, line 343: exception change here

python/tests/test_downtown_denver_example.py
```diff
@@ -97,7 +98,7 @@ def test_time(self) -> None:
 
             error = result.get("error")
             if error is not None:
-                raise Exception(
+                raise RuntimeError(
```
admrtin, line 101: exception change here

python/tests/test_downtown_denver_example.py
```diff
@@ -270,7 +271,7 @@ def test_energy(self) -> None:
...
```
admrtin, line 274: exception change here

robfitzgerald (approved):
I think this looks great. 

I learned something new: lowercase l `list` and t `tuple` are _canonical_ type representations over `List`/`Tuple` in Python. These exceptions should be `RuntimeException`s, i think that's a fine choice. i was thrown off by these explicit `display()` calls in the notebooks but I gave it a run and saw that it just makes explicit what is implicit in notebook-land and otherwise has no effect. great! 

thanks @admrtin!

PR #563, opened 2026-08-04
Pin c/c++ compilers for reproducible hpc build

nreinicke:
We ran into some build issues when moving from RHEL 8 to 9 on our hpc system caused by varying c/c++ build stacks on each of these OSes. This pins the compilers as a pixi dependnecy to guard against that happening again.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,7 +1,16 @@
 version: 7
 platforms:
 - name: linux-64
+  virtual-packages:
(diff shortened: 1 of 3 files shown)


robfitzgerald (approved):
great! thank you for figuring this one out, @nreinicke. 🎺

admrtin (approved):
The build worked perfectly with these updates. Thanks!

PR #565, opened 2026-08-07
Two hotfixes for TripHistoryTraversalEngine

admrtin:
1. TripHistoryTraversalEngine was not handling the first traversal in a route, where there is no edge to pull history from.

2. It was reading from the current state vector, not the previous.
* The history to read from is now pulled immediately as the first action in `update_history`, for clarity.


robfitzgerald (approved):
looks good. in our IRL discussion, you helped me understand that, without directly indexing state variable values during updates, we cannot further improve copying data from our previous state. thanks for all of your attention to detail here.

PR #568, opened 2026-08-13
Fix JSON array response delimiters

### LINKED ISSUE #550: JSON file output with newline_delimited = false creates rows that are not comma-delimited
when newline_delimited = false, the output is expected to be a valid json array, but prints without a comma-delimiter:

```python
[
  { 
    ...
  }    # <- missing HERE
  {
    ...
  }
]
```

Gaurav890:
## Summary

- Write the configured delimiter between file-sink responses.
- Preserve newline-delimited JSON behavior.
- Add a regression test covering multiple rows in standard JSON output.

## Root cause

`ResponseSink` stored the delimiter supplied by `ResponseOutputFormat` but ignored it, so non-newline-delimited JSON objects were written next to each other without commas.

## Validation

- `cargo test --workspace -j 1`
- `cargo fmt --all -- --check`
- `cargo clippy --all --all-targets --all-features -- -D warnings`

Fixes #550

### DIFF
--- a/rust/routee-compass/src/app/compass/response/response_sink.rs
+++ b/rust/routee-compass/src/app/compass/response/response_sink.rs
@@ -31,7 +31,7 @@ impl ResponseSink {
                 filename,
                 file,
                 format,
-                delimiter: _,
+                delimiter,
                 iterations_per_flush,
                 iterations,
             } => {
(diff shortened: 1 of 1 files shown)


robfitzgerald (approved):
looks good. thank you for providing a minimal unit test to demonstrate the working functionality. have a good day! :trumpet:

PR #570, opened 2026-08-19
Update ninterp to 0.11

kylecarow:
Made an update to ninterp, this keeps you at the state of the art. I also updated the interpolation code to take advantage of `InterpolatorEnum`, an enum directly from ninterp that does the same thing `InterpolatorVariant` did.

I ran the rust & python pixi checks and everything looked green!

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -38,7 +38,7 @@ itertools = { version = "0.15.0" }
 kdam = "0.6.3"
 log = "0.4.28"
 ndarray = "=0.17.2" # forces correct transitive dependency on ndarray by ninterp 0.8.0
-ninterp = "0.10.0"
+ninterp = "0.11.0"
 ordered-float = { version = "5.1.0", features = ["serde"] }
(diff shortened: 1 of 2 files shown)


rust/Cargo.toml
```diff
@@ -38,7 +38,7 @@ itertools = { version = "0.15.0" }
 kdam = "0.6.3"
 log = "0.4.28"
 ndarray = "=0.17.2" # forces correct transitive dependency on ndarray by ninterp 0.8.0
```
kylecarow, line 40: Note with this PR the `ndarray` version req here can probably be loosened; I switched the call site to use the re-exposed `ninterp::ndarray` for array construction.

robfitzgerald: just checked and it does appear to be the case that we can relax the import. thanks for the ergonomics improvement @kylecarow !

robfitzgerald (approved):
deeeeeeee-licious. thanks for the update, Kyle!

PR #576, opened 2026-08-27
Rjf/concurrent writer

### LINKED ISSUE #572: Slowdown caused by concurrent writes to file with many threads
for the two implementations of the File output policy, both assume synchronous writes. as we scale up to a hundred threads on HPC, this becomes a big bottleneck for our program.

to address this, we can create a file writer which still targets existing file types (CSV, JSON) but is backed by multiple files. if the order + file location of each result row does not need to align with the run order, we can maintain a pool of file pointers for writing and round-robin assign row writes to files. 

this can be implemented using a separate IO thread/pool of threads and message passing via the Rust [channel](https://doc.rust-lang.org/beta/std/sync/mpsc/fn.channel.html) type.

robfitzgerald:
Closes #572 by creating a "directory" variant of the ResponseOutputPolicy. 

## changes

- in support of both standard file and directory batch file writing, i have refactored the logic of both to use a struct called `FileState` that tracks a file pointer and the number of records written to that file. it is used by both File and Directory variants and encapsulates open/close operations that were previously inlined elsewhere.
- revised both configurations to model the output location as a PathBuf and using Path in algorithm implementations instead of String
- created constructors for ResponseSink variants pulling in logic exposed elsewhere
- retained Parquet as a format of a File output policy. while Parquet is implemented in the file system as a bunch of files, it is still referenced as a single archive, so it seemed sensible to leave it that way.

## configuration

Updated the docs/configuration.md file with the following:

#### File pools for CSV/JSON/JSONL

Writing to a pool of flat output files in CSV, JSON, or JSONL format is also possible. Similar to Parquet, concurrent writes are supported, which speeds up runtimes for large runs especially where each result contains a large amount of information.

```toml
[system.response_output_policy]
type = "directory"
path = "output/"
prefix = "data" # optional, defaults to "part"
# # sets the number of output file parts.
# # if not provided, the number of parallel threads used by Compass is used.
# concurrency = 8
format = { type = "json", newline_delimited = true }
# # other optional arguments:
# file_flush_rate = 50
# write_mode = "append" 
```


rust/routee-compass/src/app/compass/response/internal_writer.rs
```diff
@@ -83,3 +97,22 @@ fn is_empty(file: &File) -> Result<bool, CompassAppError> {
...
 }
+
+/// helper function to handle the various file type + write mode options
+fn get_or_create_file_writer(
+    filepath: &Path,
+    write_mode: &WriteMode,
+) -> Result<InternalWriter, CompassAppError> {
+    let ext_option = filepath.extension();
+    match ext_option {
+        Some(ext) if ext == "gz" => {
```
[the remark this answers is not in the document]
robfitzgerald: ...but it is compiling 😛

rust/routee-compass/src/app/compass/response/response_sink.rs
```diff
@@ -23,66 +35,151 @@ pub enum ResponseSink {
...
+        let buffer_size = file_flush_rate.unwrap_or(100) as usize;
+        let base_filename = path.to_string_lossy().to_string();
+
+        // create the parent directory if it doesn't exist
+        std::fs::create_dir(&path).map_err(|e| {
+            CompassAppError::InternalError(format!(
+                "failed to create parquet base file {:?}: {}",
+                path, e
+            ))
+        })?;
```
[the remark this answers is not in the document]
robfitzgerald: fixed

rust/routee-compass/src/app/compass/response/response_sink.rs
```diff
@@ -168,6 +271,21 @@ impl ResponseSink {
...
+    format: &ResponseOutputFormat,
+    iterations_per_flush: u64,
+) -> Result<String, CompassAppError> {
+    let mut state_attained = state.lock().map_err(|e| {
+        CompassAppError::ReadOnlyPoisonError(format!("Could not aquire lock on output file: {e}",))
+    })?;
+    if let Some(final_contents) = format.generate_footer() {
+        state_attained.write(&final_contents, None, iterations_per_flush)?;
+    }
+    Ok(state_attained.filename.clone())
```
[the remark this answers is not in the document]
robfitzgerald: good catch, fixed.

rust/routee-compass/src/app/compass/response/response_output_policy.rs
```diff
@@ -38,55 +58,44 @@ impl ResponseOutputPolicy {
...
+                };
+                ResponseSink::new_files(
+                    path.clone(),
+                    file_prefix,
+                    *gzip,
+                    *concurrency,
+                    format.clone(),
+                    *file_flush_rate,
+                    write_mode.clone(),
+                )
```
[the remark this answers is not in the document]
robfitzgerald: good catch

rust/routee-compass/src/app/compass/response/response_sink.rs
```diff
@@ -23,66 +35,151 @@ pub enum ResponseSink {
...
+        gzip: bool,
+        n_writers: Option<NonZeroUsize>,
+        format: ResponseOutputFormat,
+        file_flush_rate: Option<u64>,
+        write_mode: Option<WriteMode>,
+    ) -> Result<Self, CompassAppError> {
+        let n_writers = match n_writers {
+            Some(n) => n.into(),
+            None => rayon::current_num_threads(),
+        };
```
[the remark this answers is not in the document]
robfitzgerald: this actually already burned me, good catch.

rust/routee-compass/src/app/compass/response/response_output_policy.rs
```diff
@@ -1,30 +1,50 @@
...
 pub enum ResponseOutputPolicy {
+    /// when outputs are simply persisted by the app into program memory, such as
+    /// when using via the Rust or Python API.
     #[default]
     None,
+    /// writes all results to a single file (CSV, JSON, JSONL) or archive (Parquet).
     File {
         /// destination file. may be a standard file suffix, or, if terminates with '.gz' will be gzip-encrypted.
-        filename: String,
+        path: PathBuf,
```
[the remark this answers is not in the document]
robfitzgerald: fixed!

rust/routee-compass/src/app/compass/response/response_sink.rs
```diff
@@ -23,66 +35,151 @@ pub enum ResponseSink {
...
+                state,
+                format,
+                delimiter,
+                iterations_per_flush,
+                directory,
+            } => {
+                // map the current thread to the set of writers, wrapping around the max writer index using modulo
+                // to ensure we are always selecting a valid index, though if the user set max writers as fewer than
+                // max threads, then write operations may be blocked here as they are assigned effectively round-robin.
+                let thread_idx = rayon::current_thread_index().unwrap_or(0);
```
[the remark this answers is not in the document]
robfitzgerald: fixed

rust/routee-compass/src/app/compass/response/file_state.rs
```diff
@@ -0,0 +1,75 @@
...
+};
+
+/// the state of a file write.
+pub struct FileState {
+    /// name of the file written to
+    pub filename: String,
+    /// file writer, either a raw writer or gzip writer.
+    writer: InternalWriter,
+    /// number of records written to the file.
+    records: u64,
```
admrtin: nit: should this be num_records?

rust/routee-compass/src/app/compass/response/file_state.rs
```diff
@@ -0,0 +1,75 @@
...
+        format: &ResponseOutputFormat,
+        write_mode: Option<WriteMode>,
+    ) -> Result<Self, CompassAppError> {
+        let mut writer = InternalWriter::new(filepath, write_mode)?;
+        let filename = filepath.to_string_lossy().to_string();
+        writer.write_header(format)?;
+        Ok(FileState {
+            filename,
+            writer,
+            records: 0,
```
admrtin: very nit: could we use #[derive(Default)] and #[default] to create the writer w/0 records instead?

rust/routee-compass/src/app/compass/response/file_state.rs
```diff
@@ -0,0 +1,75 @@
...
+        if let Some(delimiter) = delimiter {
+            if self.records > 0 {
+                write!(self.writer, "{delimiter}").map_err(|e| {
+                    CompassAppError::InternalError(format!(
+                        "failure writing delimiter to {}: {e}",
+                        &self.filename
+                    ))
+                })?;
+            }
+        }
```
admrtin, line 55: just want to confirm the suggestion by copilot below isn't an issue. I can understand cases where delimiters wouldn't be needed but am moreso worried about the newline issue.

rust/routee-compass/src/app/compass/response/response_sink.rs
```diff
@@ -23,66 +35,153 @@ pub enum ResponseSink {
 }
 
 impl ResponseSink {
+    /// creates a response sink for a single file.
+    pub fn new_file(
```
admrtin, line 39: Nit: wondering if the `new_*` functions are better than having a single point of entry `pub fn new(...)`,

Since the arguments to each enum are different, the way you have it here is probably the most idiomatic, but curious if there's an even better way

rust/routee-compass/src/app/compass/response/response_output_policy.rs
admrtin, line 57: super Nit: personally, I don't love the method name "build" here, it feels a little unclear... maybe this is a standard naming convention though... i.e.  "Policy Builds Sink"? i guess i'm just trying to wrap my head around the terminology.

admrtin (changes requested):
Looks pretty good, I had some minor nits, and one go-back on a copilot suggestion.

rust/routee-compass/src/app/compass/response/file_state.rs
robfitzgerald: not here. that only works when all of the fields have known default values. however, here there is no default "filename" or "writer" that is appropriate. to support Default, FileState would need an `Option<String>` and `Option<InternalWriter>` instead.

rust/routee-compass/src/app/compass/response/file_state.rs
robfitzgerald: i ignored that comment since the proposed fix is illogical. but you're right, this FileState refactor seemed to lose knowledge of how to handle JSONL formats. i have fixed by changing `ResponseOutputFormat`. previously, if newline_delimited was true, it would set delimiter() to be None. now it will return Some("\n".to_string())

rust/routee-compass/src/app/compass/response/response_sink.rs
robfitzgerald: this is more common in Rust with enums. since the structure of the new method varies depending on the variant, it would require some `pub fn new(...)` to cover all cases and require a bunch of `Option`al arguments. but usually, if you have a variant, you usually know you want to build that variant explicitly. there are of course counter-arguments but in this case, this is the correct pattern.

rust/routee-compass/src/app/compass/response/response_output_policy.rs
robfitzgerald: here's a discussion on this topic ([link](https://users.rust-lang.org/t/rust-naming-for-creating-sub-objects/24184)).

we use the convention of a `build` method for the ~inverse~ uh, coproduct? of `new`. like, for some A that is a Factory for values of type B:
  - to make an A: `A::new()`
  - to make a B: `a.build()`

admrtin (approved):
looking good, thanks!

